PR #1, opened 2021-03-23
[MultiPlayerServers] Adding new enum state for Vms

maryammadzadeh:
This is to support manual deallocation of targeted vms


VmAgent.Core/VmAgent.Core.csproj
```diff
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
-    <PackageVersion>1.0.5</PackageVersion>
+    <PackageVersion>1.0.7</PackageVersion>
```
maryammadzadeh, line 6: https://microsoft.visualstudio.com/Xbox/_workitems/edit/32219476 Bug to track the cause for why i had to do this.

dgkanatsios (approved):

PR #8, opened 2021-04-23
Add support for empty image registry in config

jseward:
See: https://community.playfab.com/questions/52148/localmultiplayeragentexe-and-linux-container-issue.html

When testing local images there is no registry in the name. Previously the code would blindly make a corrupt imageName resulting in BadRequest:invalid reference format from docker.

Also added some convenient logging on CreateContainer to help users discover further issues on their own without stepping through source.

I'm not sure why `.Select(x => $"{x.Key}={x.Value}").ToList();` is showing up as a change... It hasn't changed.

### DIFF
--- a/VmAgent.Core/Interfaces/DockerContainerEngine.cs
+++ b/VmAgent.Core/Interfaces/DockerContainerEngine.cs
@@ -175,6 +175,7 @@ namespace Microsoft.Azure.Gaming.VmAgent.ContainerEngines
                 Cmd = startCmd
             };
 
+            _logger.LogInformation($"Creating container. Image='{imageName}'");
             CreateContainerResponse response =
                 await TaskUtil.TimedExecute(
                     async () => await _dockerClient.Containers.CreateContainerAsync(containerParams).ConfigureAwait(false),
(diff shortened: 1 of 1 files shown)


narthur157:
This worked for me, thanks!

VmAgent.Core/Interfaces/DockerContainerEngine.cs
```diff
@@ -358,7 +359,13 @@ public async Task<SessionHostInfo> CreateAndStart(int instanceNumber, GameResour
             // is the same as the remote image path.
             SessionHostsStartInfo sessionHostStartInfo = gameResourceDetails.SessionHostsStartInfo;
             ContainerImageDetails imageDetails = sessionHostStartInfo.ImageDetails;
-            string imageName = $"{imageDetails.Registry}/{imageDetails.ImageName}:{imageDetails.ImageTag ?? "latest"}";
+            string imageName = $"{imageDetails.ImageName}:{imageDetails.ImageTag ?? "latest"}";
```
dgkanatsios, line 362: thanks for the PR! Can you wrap the change in a `if sessionHostManager.LinuxContainersOnWindows` clause, to make sure this applies only on LocalMultiplayerAgent and not on VmAgent running on production VMs? I don't think this will have a negative impact there, but I'd rather not change existing behavior unless it's absolutely necessary. I'll also make a note on the contribution notes after the PR that changes in the .Core project that target LocalMultiplayerAgent should make use of this flag. Thanks again!

dgkanatsios (changes requested):

VmAgent.Core/Interfaces/DockerContainerEngine.cs
jseward: Hi @dgkanatsios , I don't mind adding that at all but is it the right guard? Was just luck that I didn't run into this issue when testing window containers due to the image names.

dgkanatsios: got it, you're right. We're good, thanks!

dgkanatsios (approved):

PR #21, opened 2021-06-17
[Update] Adding support for new enum to indicate there are active sessions

maryammadzadeh:
This enum will be used to determine the action the session contoller in the agent will take when sending back responses for unassignment.

### DIFF
--- a/AgentInterfaces/VmState.cs
+++ b/AgentInterfaces/VmState.cs
@@ -101,6 +101,12 @@ namespace Microsoft.Azure.Gaming.AgentInterfaces
         /// The game servers are exiting too quickly (potentially due to crashes).
         /// </summary>
         [ProtoEnum]
-        TooManyServerRestarts
+        TooManyServerRestarts,
+
+        /// <summary>
+        /// There are game servers with active sessions
+        /// </summary>
+        [ProtoEnum]
+        HasActiveSessions
     }
 }


AgentInterfaces/VmState.cs
```diff
@@ -101,6 +101,12 @@ public enum VmState
...
         /// </summary>
         [ProtoEnum]
-        TooManyServerRestarts
+        TooManyServerRestarts,
+
+        /// <summary>
+        /// There are game servers with active sessions
+        /// </summary>
+        [ProtoEnum]
+        HasActiveSessions
```
ravarnamsft, line 110: I am not sure if this is really needed. The unassign API on the VM Agent checks that there are no active sessions (and rejects the unassignment call). Moreover, our states are not flags and cannot be combined. How do we combine this with other states potentially? Please check the unassign codepath.

maryammadzadeh: We collapse all none 200s and retry indefinitely, so i would like to add clearer logic there. From cp, i went to mps agent to see how the controller returns the results and followed that to how we check for isunassigned. 

My plan is to check for this specific state and then return the this state as it makes the code understandable. 

We could use Assigned, too.

maryammadzadeh: I would actually like to return the count of active sessions so we know, too. Do you know if there was a reason to make the request just use the VERB and result, and no content, For perf?

maryammadzadeh: The other way to solve this is to to keep a dictionary of the vm we are trying to unassign in cp and if we get a timeout from vm agent, we can just mark it 4 times and then officially call unassign.

But if cp is getting timeouts but not because of vmagents response not arriving (due to other perf issues) then it might assume that the timeouts are legit and mark the vm and we'd still be in the same place.

maryammadzadeh: The other way to solve this is to to keep a dictionary of the vms we are trying to unassign in cp and if we get a timeout from vm agent, we can just mark it 4 times and then officially call unassign.

ravarnamsft: The heartbeats of the VM already has the count of actives etc right? When you are trying to unassign, all we care from the VMAgent is whether it has active sessions (if so reject), if not accept it and complete the operation. As for indefinite tries, I thought Hesky added some timeouts to the unassign part (which may not be enabled right now, something to look at). CP has clear codes for unassignment (UnassignResult) that can distinguish what's happening. 

"The other way to solve this is to to keep a dictionary of the vms we are trying to unassign in cp and if we get a timeout from vm agent, we can just mark it 4 times and then officially call unassign." - I think this is the part that should be in place but may not be exercised. Tagging @andhesky  since he had worked on timeouts (it was mostly for Assign, but I think it works for Unassign as well).

Knowing the count of active sessions was not needed for the call, we can just edit the error message that's returned as part of the 409 conflict to say how many sessions are there, if we care about it for debugging. 
The following line in SessionHostManager (line 346) in Vmagent just says one or more. If we need (not saying we shoud), we can add the number there:
_throw new InvalidUnassignmentException("The VM has one more active game servers.");_

This work item is a bit nuanced, so it would help to see what your full plan to tackle it is (it is ok to do it in phases, but it is hard to sign off on individual pieces without fully knowing how the solution will look like).

maryammadzadeh: I'm totally ok with keep the changes only in CP, and it makes a lot more sense to keep the log in the exception, Thanks.

PR #22, opened 2021-06-18
added VmAgentSettings class

brianwp3000:
This class will allow VmAgent to request settings from ControlPlane. I have a separate pull request in ControlPlane that shows how this class will be used.

### DIFF
--- a/AgentInterfaces/VmAgentSettings.cs
+++ b/AgentInterfaces/VmAgentSettings.cs
@@ -5,8 +5,8 @@ namespace Microsoft.Azure.Gaming.AgentInterfaces
 {
     public class VmAgentSettings
     {
-        public bool ProcessCrashDumps { get; set; }
+        public bool EnableCrashDumpProcessing { get; set; }
 
-        public bool SendTelemetry { get; set; }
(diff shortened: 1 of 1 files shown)


AgentInterfaces/VmAgentSettings.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Azure.Gaming.AgentInterfaces
+{
+    public class VmAgentSettings
+    {
+        public bool ProcessCrashDumps { get; set; }
```
maryammadzadeh, line 8: Boolean should be a question, what about ShouldEnableTelemetry and ShouldEnableCrashDumps?

maryammadzadeh (approved):

AgentInterfaces/VmAgentSettings.cs
brianwp3000: changed to `EnableCrashDumpProcessing` and `EnableTelemetry`

PR #26, opened 2021-06-25
added VmAgentSettings to ISessionHostManager

brianwp3000:
DockerContainerEngine and ProcessRunner show how the new VmAgentSettings will be used for dump collection in the future.
I need this checked in so i can put it in the VmAgent.Core nuget package and get it into our internal VmAgent project.

### DIFF
--- a/LocalMultiplayerAgent/NoOpSessionHostManager.cs
+++ b/LocalMultiplayerAgent/NoOpSessionHostManager.cs
@@ -29,8 +29,8 @@ namespace Microsoft.Azure.Gaming.LocalMultiplayerAgent
 
         public VmAgentSettings VmAgentSettings { get; } = new VmAgentSettings() 
         { 
-            EnableCrashDumpProcessing = true,
-            EnableTelemetry = true,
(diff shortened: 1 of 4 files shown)


LocalMultiplayerAgent/NoOpSessionHostManager.cs
```diff
@@ -27,6 +27,12 @@ public bool LinuxContainersOnWindows
 
         public int StateChangeSequenceNumber { get; }
 
+        public VmAgentSettings VmAgentSettings { get; } = new VmAgentSettings() 
+        { 
+            EnableCrashDumpProcessing = true,
```
dgkanatsios, line 32: should we have them as false?

brianwp3000: My thinking was that we want these settings to all work in LocalMultiplayerAgent as well, but i can see it being worth leaving them false so that we aren't making developers alpha test for us.

dgkanatsios (approved):

PR #30, opened 2021-07-09
Example for running on MacOs

crazedpeanut:
Hi! I understand MacOS isn't officially supported but I thought i'd share the changes I needed to make in order to run the agent on my MacBook.

I had to make the following changes to the configuration file:

- Disable RunContainer
- The StartGameCommand needed to include the path of the actual binary for the game, not the package. 
For example: 
"ProcessStartParameters": {
        "StartGameCommand": "/Users/example/Example.app/Contents/MacOS/Example -nographics -batchmode"
    },

Our Unity Game's product name includes spaces, so I needed to escape these in the ProcessRunner. Otherwise, the StartGameCommand would be interpreted incorrectly.

It works great, thanks for providing the local agent!

### DIFF
--- a/LocalMultiplayerAgent/Config/MultiplayerSettingsValidator.cs
+++ b/LocalMultiplayerAgent/Config/MultiplayerSettingsValidator.cs
@@ -33,17 +33,17 @@ namespace Microsoft.Azure.Gaming.LocalMultiplayerAgent.Config
                 throw new Exception("OutputFolder or TitleId not found. Call SetDefaultsIfNotSpecified() before this method");
             }
 
-            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
-            {
-                Console.WriteLine("Running LocalMultiplayerAgent on Linux is not supported yet.");
-                return false;
-            }
-
-            if (Globals.GameServerEnvironment == GameServerEnvironment.Linux && !_settings.RunContainer)
-            {
-                Console.WriteLine("The specified settings are invalid. Using Linux Game Servers requires running in a container.");
-                return false;
-            }
+            // if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            // {
+            //     Console.WriteLine("Running LocalMultiplayerAgent on Linux is not supported yet.");
+            //     return false;
+            // }
+            //
+            // if (Globals.GameServerEnvironment == GameServerEnvironment.Linux && !_settings.RunContainer)
+            // {
+            //     Console.WriteLine("The specified settings are invalid. Using Linux Game Servers requires running in a container.");
+            //     return false;
+            // }
 
             string startGameCommand;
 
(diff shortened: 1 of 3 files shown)


LocalMultiplayerAgent/Config/MultiplayerSettingsValidator.cs
```diff
@@ -33,17 +33,17 @@ public bool IsValid()
...
+            // {
+            //     Console.WriteLine("Running LocalMultiplayerAgent on Linux is not supported yet.");
+            //     return false;
+            // }
+            //
+            // if (Globals.GameServerEnvironment == GameServerEnvironment.Linux && !_settings.RunContainer)
+            // {
+            //     Console.WriteLine("The specified settings are invalid. Using Linux Game Servers requires running in a container.");
+            //     return false;
+            // }
```
maryammadzadeh, line 46: Thanks for contributing! It seems that we should have some conditional logic in place as well to see if we are running in macos to prevent the check, rather than comment it out.  Similarly for the ProcessRunner start command. @dgkanatsios  thoughts?

crazedpeanut: I'm happy to push another tidier commit. I wasn't sure if you'd be interested in MacOS support!

I can also add an example config

dgkanatsios:
hey @crazedpeanut, wow! thanks a lot for this, we'd love to have MacOS support!!! We'd be more than happy to work with you on this PR :) May I ask, did you try running container mode as well? Any luck?

crazedpeanut:
@dgkanatsios I didn't try running it in a container yet. 

Thinking about it now I probably should have tried running a Linux container first. I probably wouldn't have had to change anything!

I'll have a look

dgkanatsios:
@crazedpeanut thank you! So my current thinking is that developers using MacOS would probably target Linux as the operating system of choice of PlayFab. In this context, I think that LocalMultiplayerAgent should support containers on Linux when running on MacOS. Let me know what you think, thanks again for working on this!

crazedpeanut:
Yes that sounds right! I’ll test out running in a Linux container

crazedpeanut:
@dgkanatsios it turns out the project I am working on can't target Linux because of some dependency issues. So I am stuck with the MacOS builds. 

If this is not the intended workflow for the local agent feel free to close this issue, i'll just keep my own fork :)

dgkanatsios:
@crazedpeanut so the way it works right now is launching MacOS processes?

dgkanatsios:
Great! So we could work and eventually merge this, by having a check in the beginning like “if OS==MacOS and runContainers==true then exit”. What do you think? We’d be more than happy to accept it!

VmAgent.Core/Interfaces/ProcessRunner.cs
```diff
@@ -92,7 +92,11 @@ public Task CollectLogs(string id, string logsFolder, ISessionHostManager sessio
 
         private (string, string) GetExecutableAndArguments(SessionHostsStartInfo sessionHostsStartInfo, int instanceNumber)
         {
-            string[] parts = sessionHostsStartInfo.StartGameCommand.Split(' ', StringSplitOptions.RemoveEmptyEntries);
+            var command = sessionHostsStartInfo.StartGameCommand.Replace("\\ ", "\\SPACE");
```
pgilmorepf, line 95: C# generally accepts / in Windows paths, and Linux-based systems naturally want / as well. I'd say a conversion of "\\" -> "/" would be more universal and safe. Additionally, when I think of what would typically replace a space in a path, I think of "_" or sometimes "-". I don't really think of "SPACE".

PR #31, opened 2021-07-15
Bport/create dump folder

brianwp3000:
On startup, creates a `_dumps` folder inside of the logs folder and creates a `PF_SERVER_DUMPS_DIRECTORY` environment variable that points to it. 
On exit, deletes the `_dumps` folder if it's empty (this way, developers won't have an empty folder in all of their log archives). If it's not empty, a call is made to the `ISessionHostManager` to signal that a dump was found and to determine if dump processing has been throttled. If it's been throttled, it deletes the dump and replaces it with a `readme.txt` file with a short explanation.
LocalMultiplayerAgent uses a `NoopSessionHostManager`, which doesn't do any actual dump signaling or throttling. That logic will be implemented in our internal VmAgent repo.

All of these changes are put behind the EnableCrashDumpProcessing setting.

### DIFF
--- a/VmAgent.Core/Interfaces/BaseSessionHostRunner.cs
+++ b/VmAgent.Core/Interfaces/BaseSessionHostRunner.cs
@@ -11,6 +11,7 @@ namespace Microsoft.Azure.Gaming.VmAgent.ContainerEngines
     using Microsoft.Azure.Gaming.AgentInterfaces;
     using Microsoft.Azure.Gaming.VmAgent.Core;
     using System.IO;
+    using System;
 
     public abstract class BaseSessionHostRunner : ISessionHostRunner
     {
@@ -86,6 +87,10 @@ namespace Microsoft.Azure.Gaming.VmAgent.ContainerEngines
                     // in the dumps folder, and then the game server crashed. The background process could theoretically still be
                     // running, which would prevent us from processing the dump files.
                     _logger.LogWarning($"Unable to process dump files on session host {id}: {ex}");
+                }
+                catch (Exception ex)
+                {
+                    _logger.LogError($"Unexpected error processing dump files on session host {id}: {ex}");
                 }
             }
         }
--- a/VmAgent.Core/Interfaces/BaseSessionHostRunner.cs
+++ b/VmAgent.Core/Interfaces/BaseSessionHostRunner.cs
@@ -85,7 +85,7 @@ namespace Microsoft.Azure.Gaming.VmAgent.ContainerEngines
                     // I think we'd only end up here if a game server spun up a background process that had a lock on some files
                     // in the dumps folder, and then the game server crashed. The background process could theoretically still be
                     // running, which would prevent us from processing the dump files.
-                    _logger.LogWarning($"Unable to process dump files: {ex}");
+                    _logger.LogWarning($"Unable to process dump files on session host {id}: {ex}");
                 }
             }
         }
--- a/VmAgent.Core/Interfaces/BaseSessionHostRunner.cs
+++ b/VmAgent.Core/Interfaces/BaseSessionHostRunner.cs
@@ -1,0 +1,93 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Azure.Gaming.VmAgent.ContainerEngines
+{
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.Gaming.VmAgent.Model;
+    using Core.Interfaces;
+
+    using Microsoft.Azure.Gaming.AgentInterfaces;
+    using Microsoft.Azure.Gaming.VmAgent.Core;
+    using System.IO;
+
+    public abstract class BaseSessionHostRunner : ISessionHostRunner
+    {
+        protected readonly MultiLogger _logger;
+
+        protected readonly VmConfiguration _vmConfiguration;
+
+        protected readonly ISystemOperations _systemOperations;
+
+        protected BaseSessionHostRunner(
+            VmConfiguration vmConfiguration,
+            MultiLogger logger,
+            ISystemOperations systemOperations)
+        {
+            _logger = logger;
+            _vmConfiguration = vmConfiguration;
+            _systemOperations = systemOperations;
+        }
+
+        abstract public Task CollectLogs(string id, string logsFolder, ISessionHostManager sessionHostManager);
+
+        abstract public Task<SessionHostInfo> CreateAndStart(int instanceNumber, GameResourceDetails gameResourceDetails, ISessionHostManager sessionHostManager);
+
+        abstract public Task DeleteResources(SessionHostsStartInfo sessionHostsStartInfo);
+
+        abstract public string GetVmAgentIpAddress();
+
+        abstract public Task<IEnumerable<string>> List();
+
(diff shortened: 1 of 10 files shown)


VmAgent.Core/Interfaces/ProcessRunner.cs
```diff
@@ -84,7 +91,43 @@ public Task CollectLogs(string id, string logsFolder, ISessionHostManager sessio
...
+                        }
+                    }
+                    catch (DirectoryNotFoundException) { }
+
+                    if (dumpFound)
+                    {
+                        bool shouldDeleteDump = sessionHostManager.SignalDumpFoundAndCheckIfThrottled(id);
+                        if (shouldDeleteDump)
+                        {
+                            _systemOperations.DeleteDirectoryIfExists(dumpFolder);
```
ravarnamsft, line 117: the code looks very familiar to the one for containers. I understand we do not have a base class but perhaps we should move this to a common helper class?

VmAgent.Core/Interfaces/ISessionHostManager.cs
```diff
@@ -70,5 +70,12 @@ public interface ISessionHostManager
         bool IsStartupScriptExecutionComplete();
 
         bool IsUnassignable();
+
+        /// <summary>
+        /// Signals that a crash dump was found in the given SessionHost's logs folder, and checks
+        /// if the crash dump should be deleted due to throttling.
+        /// </summary>
+        /// <returns>True if the crash dump should be deleted, false otherwise</returns>
+        bool SignalDumpFoundAndCheckIfThrottled(string sessionHostId);
```
ravarnamsft, line 79: Can we name it "ShouldDeleteDumpsFromLogsFolder". I am guessing it will only be called when we find a dump file, so the signal part is potentially redundant.

ravarnamsft (approved):
:shipit:

VmAgent.Core/Interfaces/ISessionHostManager.cs
brianwp3000: I'm not happy with the current name either, but my thinking was that this method will be where we put the logic that puts the necessary info into the heartbeat for a playstream event to get fired. So this method will "signal that a dump was found". I actually did consider naming it "ShouldDeleteDumps" but that doesn't capture the fact that a signal will be sent, which might be an unintended side effect depending on where it's called.

VmAgent.Core/Interfaces/ProcessRunner.cs
brianwp3000: i ended up creating a base class for them and moved this logic into the ProcessDumps method there

VmAgent.Core/Interfaces/BaseSessionHostRunner.cs
```diff
@@ -0,0 +1,93 @@
...
+                            _systemOperations.FileWriteAllText(readmePath, $"The contents of \"{VmDirectories.GameDumpsFolderName}\" have been deleted due to throttling.");
+                        }
+                    }
+                }
+                catch (IOException ex)
+                {
+                    // I think we'd only end up here if a game server spun up a background process that had a lock on some files
+                    // in the dumps folder, and then the game server crashed. The background process could theoretically still be
+                    // running, which would prevent us from processing the dump files.
+                    _logger.LogWarning($"Unable to process dump files: {ex}");
```
maryammadzadeh, line 88: Log the id,  and log folder?

VmAgent.Core/Interfaces/BaseSessionHostRunner.cs
```diff
@@ -0,0 +1,93 @@
...
+                        {
+                            dumpFound = true;
+                        }
+                        else
+                        {
+                            // If dumps folder is empty, delete it
+                            _systemOperations.DeleteDirectoryIfExists(dumpFolder);
+                        }
+                    }
+                    catch (DirectoryNotFoundException) { }
```
maryammadzadeh, line 69: Do we need to log here?

VmAgent.Core/Interfaces/BaseSessionHostRunner.cs
```diff
@@ -0,0 +1,93 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
```
maryammadzadeh, line 2: Not related to your PR, never noticed MIT in this codebase

VmAgent.Core/Interfaces/BaseSessionHostRunner.cs
brianwp3000: Added session host id (good catch--i assumed that'd get logged by default but it isn't), but there isn't much value in logging the folder path--the folder gets deleted after the logs are uploaded so we wouldn't be able to navigate to it or anything

VmAgent.Core/Interfaces/BaseSessionHostRunner.cs
```diff
@@ -0,0 +1,93 @@
...
+                        }
+                    }
+                }
+                catch (IOException ex)
+                {
+                    // I think we'd only end up here if a game server spun up a background process that had a lock on some files
+                    // in the dumps folder, and then the game server crashed. The background process could theoretically still be
+                    // running, which would prevent us from processing the dump files.
+                    _logger.LogWarning($"Unable to process dump files on session host {id}: {ex}");
+                }
```
maryammadzadeh, line 89: Should we catch and log any other type of exception here just in case? At least catch Exception.

VmAgent.Core/Interfaces/DockerContainerEngine.cs
```diff
@@ -58,11 +51,9 @@ public class DockerContainerEngine : ISessionHostRunner
             MultiLogger logger,
             Core.Interfaces.ISystemOperations systemOperations,
             IDockerClient dockerClient = null)
+            : base (vmConfiguration, logger, systemOperations)
         {
             // This can be moved to dependency injection pattern when unit tests are added for this class.
```
maryammadzadeh, line 56: Do we need this comment anymore?

VmAgent.Core/Interfaces/SystemOperations.cs
```diff
@@ -68,6 +69,11 @@ public void DeleteDirectoryIfExists(string pathToDirectory)
             }
         }
 
+        public bool IsDirectoryEmpty(string pathToDirectory)
+        {
+            return Directory.EnumerateFileSystemEntries(pathToDirectory).Any();
```
maryammadzadeh, line 74: Any() returns trye if there are any. Should we rename this method or return the opposite?

maryammadzadeh (changes requested):
Looks good, have comments on the name of a method and exception handling.

VmAgent.Core/Interfaces/BaseSessionHostRunner.cs
brianwp3000: Added catch(Exception) with a LogError

VmAgent.Core/Interfaces/DockerContainerEngine.cs
brianwp3000: removed

VmAgent.Core/Interfaces/SystemOperations.cs
brianwp3000: Yep good catch

maryammadzadeh (approved):

PR #39, opened 2021-07-27
Validator update

### LINKED ISSUE #37: Missing validation check for NodePort when running in container mode
When RunContainer is true, should check for the NodePort to not be null/empty

valexao:
closes #37


LocalMultiplayerAgent/Config/MultiplayerSettings.cs
```diff
@@ -46,6 +46,8 @@ public class MultiplayerSettings
         public ProcessStartParameters ProcessStartParameters { get; set; }
 
         public SessionConfig SessionConfig { get; set; }
+
+        public int NodePort { get; set; }
```
dgkanatsios, line 50: NodePort is a property in the array PortMappingsList, specifically on the PortMapping object. That property is the one we need to validate, right?

dgkanatsios (approved):
:shipit:

PR #41, opened 2021-07-28
Add timebounding for running the monitor application

IvanLH:


AgentInterfaces/SessionHostsStartInfo.cs
```diff
@@ -167,21 +167,27 @@ public class MonitoringApplicationConfiguration
...
+        public string ExecutionScript { get; set; }
 
         /// <summary>
         /// Indicates whether the monitoring application should be run in on demand mode or from the beginning 
         /// </summary>
-        public bool MonitoringApplicationRunOnDemand { get; set; }
+        public bool RunOnDemand { get; set; }
+
+        /// <summary>
+        /// If the monitoring application is to be run from the beggining, for how long should it run.
```
dgkanatsios, line 188: is this defined in hours/minutes/seconds? Could it be a Timespan?

dgkanatsios: Also, how is it going to be triggered if not run from the beginning?

IvanLH: Fixed the name, I want -1 to mean no limit, and TimeSpan would only have 0 which is a bit less clear IMO.

If the thing is not run from the beginning it will run until the caller server ends, or will be timebount in the GSDK call.

VmAgent.Core/VmAgent.Core.csproj
```diff
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
-    <PackageVersion>1.1.8</PackageVersion>
+    <PackageVersion>1.1.9</PackageVersion>
```
dgkanatsios, line 6: do we need to increase the version number here?

dgkanatsios (approved):
:shipit:

VmAgent.Core/VmAgent.Core.csproj
IvanLH: So this packages AgentInterfaces along with VMCore, VM Agent only consumes VmCore, not AgentInterfaces separately. Whereas CP only consumes AI. Thus a change in AI needs a core version bump :)

PR #43, opened 2021-07-31
Further needed changes to support profiling

IvanLH:


dgkanatsios (approved):

AgentInterfaces/VmAgentInfo.cs
```diff
@@ -47,6 +47,12 @@ public class VmAgentInfo
         [ProtoMember(8)]
         public VmNetworkConfiguration NetworkConfiguration { get; set; }
 
+        /// <summary>
+        /// Monitoring work product for this VM
+        /// </summary>
+        [ProtoMember(9)]
+        public string VmMonitoringOutput { get; set; }
```
ravarnamsft, line 54: Sorry, I have been out of the loop for a bit, what does this look like and how is it used in CP?

VmAgent.Core/Model/MonitoringPersistedState.cs
```diff
@@ -12,13 +12,15 @@ public enum MonitoringApplicationState
...
         ExecutionComplete
     }
     public class MonitoringPersistedState
     {
         public int MonitoringPID { get; set; }
-        public int RetryCount { get; set; }
         public MonitoringApplicationState MonitoringState { get; set; }
-        public DateTime ApplicationStartTime { get; set; } 
+        public DateTime ExpectedCompletionTime { get; set; } 
+        public string CallerSessionID { get; set; }
```
ravarnamsft, line 23: Minor: I think we have been using "Id" instead of "ID" in most places. Consider updating this and MonitoringPID above

ravarnamsft:
>         Downloaded,

Aren't these part of the regular asset download?

---
Refers to: VmAgent.Core/Model/MonitoringPersistedState.cs:9 in 539682e. [](commit_id = 539682e59256b4f94e1fcca4341483b648dca38c, deletion_comment = False)

PR #44, opened 2021-08-04
removing crash dump processing logic from this solution...

brianwp3000:
...it's going to be in VmAgent sln. Also added CrashDumpState to SessionHostHeartbeatInfo

### DIFF
--- a/AgentInterfaces/AgentInterfaces.csproj
+++ b/AgentInterfaces/AgentInterfaces.csproj
@@ -9,7 +9,7 @@
     -->
     <GeneratePackageOnBuild>True</GeneratePackageOnBuild>
     <PackageId>PlayFab.MultiplayerServers.AgentInterfaces</PackageId>
-    <PackageVersion>5.1.5</PackageVersion>
+    <PackageVersion>5.2.0</PackageVersion>
     <PackageDescription>Helper library for LocalMultiplayerAgent, part of Azure PlayFab Multiplayer Servers </PackageDescription>
     <GenerateAssemblyInfo>False</GenerateAssemblyInfo>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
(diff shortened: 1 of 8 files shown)


AgentInterfaces/AgentInterfaces.csproj
```diff
@@ -9,7 +9,7 @@
     -->
     <GeneratePackageOnBuild>True</GeneratePackageOnBuild>
     <PackageId>PlayFab.MultiplayerServers.AgentInterfaces</PackageId>
-    <PackageVersion>5.1.5</PackageVersion>
+    <PackageVersion>5.2.0</PackageVersion>
```
dgkanatsios, line 12: why not 5.1.6? :)

VmAgent.Core/VmAgent.Core.csproj
```diff
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
-    <PackageVersion>1.1.10</PackageVersion>
+    <PackageVersion>1.2.0</PackageVersion>
```
dgkanatsios, line 6: why not 1.1.11? :)

dgkanatsios (approved):

VmAgent.Core/VmAgent.Core.csproj
brianwp3000: This change should technically be a major version bump--i removed a method from a public-facing interface, which is a breaking change. I opted for bumping the minor version instead though because it was a method that wasn't in use yet.
I'd prefer bumping the major version over bumping the patch version, though, because changing the patch version implies that there is no change to the public API. We should only be bumping the patch version if it's an implementation change that doesn't affect compilation.

AgentInterfaces/AgentInterfaces.csproj
brianwp3000: This change is the definition of a minor version change--it's a backwards-compatible change to the public API (i.e. i added a field to SessionHostHeartbeatInfo)

VmAgent.Core/VmAgent.Core.csproj
dgkanatsios: fair enough, though our internal code is the only consumer of these APIs. LGTM!

PR #49, opened 2021-08-11
couple of bugs

dgkanatsios:
1. on Process Mode, if ContainerStart parameters are missing, LocalMultiplayerAgent will crash
2. making the _logger on SystemOperations run only if != null (which is the case on LocalMultiplayerAgent)


brianwp3000 (approved):

dgkanatsios:
Note that we don't bump VmAgent.Core version since @IvanLH is working on this project now

PR #59, opened 2021-08-20
Added WindowsCrashDumpConfiguration to SessionHostsStartInfo

brianwp3000:
Now ControlPlane can send the values needed when the vm is assigned so that VmAgent can set the appropriate registry keys whenever it creates a Windows Container. When set, these keys will cause crash dumps to automatically get collected and placed in the _dumps folder.

### DIFF
--- a/VmAgent.Core/Interfaces/DockerContainerEngine.cs
+++ b/VmAgent.Core/Interfaces/DockerContainerEngine.cs
@@ -603,7 +603,22 @@ namespace Microsoft.Azure.Gaming.VmAgent.ContainerEngines
             {
                 if (_systemOperations.IsOSPlatform(OSPlatform.Windows))
                 {
-                    return new List<string>() { $"cmd /c {request.StartGameCommand}" };
+                    var windowsStartCommand = new List<string>();
+                    
+                    if (request.WindowsCrashDumpConfiguration?.IsEnabled == true)
+                    {
+                        // set crash dump registry keys on startup
+                        windowsStartCommand.AddRange(new string[] {
+                            $"cmd /c reg add \"HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\Windows Error Reporting\\LocalDumps\" /v DumpFolder /t REG_EXPAND_SZ /d %PF_SERVER_DUMP_DIRECTORY% /f ;",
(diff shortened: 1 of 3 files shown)


VmAgent.Core/Interfaces/DockerContainerEngine.cs
```diff
@@ -603,7 +603,22 @@ private IList<string> GetStartGameCmd(SessionHostsStartInfo request)
...
+                    var windowsStartCommand = new List<string>();
+                    
+                    if (request.WindowsCrashDumpConfiguration?.IsEnabled == true)
+                    {
+                        // set crash dump registry keys on startup
+                        windowsStartCommand.AddRange(new string[] {
+                            $"cmd /c reg add \"HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\Windows Error Reporting\\LocalDumps\" /v DumpFolder /t REG_EXPAND_SZ /d %PF_SERVER_DUMP_DIRECTORY% /f ;",
+                            $"cmd /c reg add \"HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\Windows Error Reporting\\LocalDumps\" /v DumpCount /t REG_DWORD /d 10 /f ;",
+                            $"cmd /c reg add \"HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\Windows Error Reporting\\LocalDumps\" /v DumpType /t REG_DWORD /d {request.WindowsCrashDumpConfiguration.DumpType} /f ;",
+                            $"cmd /c reg add \"HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\Windows Error Reporting\\LocalDumps\" /v CustomDumpFlags /t REG_DWORD /d {request.WindowsCrashDumpConfiguration.CustomDumpFlags} /f ;",
```
maryammadzadeh, line 615: Will the request values always be set? Should we have default? Is it ok if its null?

brianwp3000: Those are both ints, so as long as `WindowsCrashDumpConfiguration` isn't null (which we check for) then they will at least have a value. And while 0 is a value value for both of these registry keys so that's not an issue. I think validation at build creation time should be sufficient

maryammadzadeh (approved):
:shipit:

Issue #48, opened 2021-08-11
LocalMultiplayerAgent should validate StartGameCommand


dgkanatsios:
LocalMultiplayerAgent should validate StartGameCommand both for processes/containers on Windows/Linux.

1. It should check if the StartGameCommand starts with the MountPath (for container mode). 
2. It should check if the executable/script of the StartGameCommand exists in the assets .zip file

valexao:
The second goal to check if the executable/script exists in the assets .zip file is still open. I believe that a solution could be to use ZipArchive's .GetEntry() to search for the specific file. There are many cases where users will have other command line instructions in their start game command, so parsing the start game command is not straight forward.

PR #64, opened 2021-09-16
Start Game Command validation

valexao:
solves the first part of issue #48 and validates that the start game command contains the mount path.


LocalMultiplayerAgent.UnitTest/ConfigurationTests.cs
```diff
@@ -250,5 +250,20 @@ public void EmptyNodePortInContainerModeShouldFail()
             new MultiplayerSettingsValidator(settings, _mockSystemOperations.Object).IsValid().Should().BeFalse();
 
         }
+
+        [TestMethod]
```
dgkanatsios, line 254: can you add a couple more test cases? e.g. a test that succeeds and multiple variations of startgamecommand.

dgkanatsios (approved):
:shipit:

PR #78, opened 2021-12-02
Support running Linux Process

emmayspark:
To support running Linux Process on Linux in LocalMultiplayerAgent
1) deleted OS validation code in MultiplayerSettingsValidator.cs.
2) added file extension variables in Globals.cs - these variables are also defined in VmAgent
3) added GetProcessStartInfoForZip() and GetProcessStartInfoForTarOrGZip to support zip and tar/targ on Linux (defined in VmAgent)

### DIFF
--- a/LocalMultiplayerAgent/Config/MultiplayerSettingsValidator.cs
+++ b/LocalMultiplayerAgent/Config/MultiplayerSettingsValidator.cs
@@ -37,7 +37,7 @@ namespace Microsoft.Azure.Gaming.LocalMultiplayerAgent.Config
 
             if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
             {
-                Console.WriteLine("Running LocalMultiplayerAgent is not yet supported on MacOS.");
+                Console.WriteLine("Running LocalMultiplayerAgent is not yet supported on MacOS. We would be happy to accept PRs to make it work!");
                 return false;
             }
 
--- a/LocalMultiplayerAgent/Config/MultiplayerSettingsValidator.cs
+++ b/LocalMultiplayerAgent/Config/MultiplayerSettingsValidator.cs
@@ -35,9 +35,15 @@ namespace Microsoft.Azure.Gaming.LocalMultiplayerAgent.Config
                 throw new Exception("OutputFolder or TitleId not found. Call SetDefaultsIfNotSpecified() before this method");
             }
 
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                Console.WriteLine("Running LocalMultiplayerAgent is not yet supported on MacOS.");
+                return false;
+            }
+
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux) && _settings.RunContainer)
             {
-                Console.WriteLine("Running LocalMultiplayerAgent as container mode on Linux is not supported yet.");
+                Console.WriteLine("Running LocalMultiplayerAgent as container mode is not yet supported on Linux. Please set RunContainer to false in MultiplayerSettings.json");
                 return false;
             }
 
--- a/LocalMultiplayerAgent/Config/MultiplayerSettingsValidator.cs
+++ b/LocalMultiplayerAgent/Config/MultiplayerSettingsValidator.cs
@@ -35,11 +35,11 @@ namespace Microsoft.Azure.Gaming.LocalMultiplayerAgent.Config
                 throw new Exception("OutputFolder or TitleId not found. Call SetDefaultsIfNotSpecified() before this method");
             }
 
-            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
-            {
-                Console.WriteLine("Running LocalMultiplayerAgent on Linux is not supported yet.");
-                return false;
-            }
+            //if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            //{
+            //    Console.WriteLine("Running LocalMultiplayerAgent on Linux is not supported yet.");
+            //    return false;
+            //}
 
             if (Globals.GameServerEnvironment == GameServerEnvironment.Linux && !_settings.RunContainer)
             {
--- a/LocalMultiplayerAgent/Config/MultiplayerSettingsValidator.cs
+++ b/LocalMultiplayerAgent/Config/MultiplayerSettingsValidator.cs
@@ -35,11 +35,11 @@ namespace Microsoft.Azure.Gaming.LocalMultiplayerAgent.Config
                 throw new Exception("OutputFolder or TitleId not found. Call SetDefaultsIfNotSpecified() before this method");
             }
 
-            //if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
-            //{
-            //    Console.WriteLine("Running LocalMultiplayerAgent on Linux is not supported yet.");
-            //    return false;
-            //}
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux) && _settings.RunContainer)
+            {
+                Console.WriteLine("Running LocalMultiplayerAgent as container mode on Linux is not supported yet.");
+                return false;
+            }
 
             if (Globals.GameServerEnvironment == GameServerEnvironment.Linux && !_settings.RunContainer)
             {
--- a/LocalMultiplayerAgent/Globals.cs
+++ b/LocalMultiplayerAgent/Globals.cs
@@ -29,12 +29,6 @@ namespace Microsoft.Azure.Gaming.LocalMultiplayerAgent
             new MultiLogger(Logger, new TelemetryClient(TelemetryConfiguration.CreateDefault()));
 
         public static GameServerEnvironment GameServerEnvironment { get; set; }
-
-        public const string ZipExtension = ".zip";
-
-        public const string TarExtension = ".tar";
-
-        public const string TarGZipExtension = ".tar.gz";
     }
 
     public enum GameServerEnvironment
--- a/LocalMultiplayerAgent/Globals.cs
+++ b/LocalMultiplayerAgent/Globals.cs
@@ -30,6 +30,11 @@ namespace Microsoft.Azure.Gaming.LocalMultiplayerAgent
 
         public static GameServerEnvironment GameServerEnvironment { get; set; }
 
+        public const string ZipExtension = ".zip";
+
+        public const string TarExtension = ".tar";
+
+        public const string TarGZipExtension = ".tar.gz";
     }
 
     public enum GameServerEnvironment
--- a/LocalMultiplayerAgent/MultiplayerServerManager.cs
+++ b/LocalMultiplayerAgent/MultiplayerServerManager.cs
@@ -32,13 +32,13 @@ namespace Microsoft.Azure.Gaming.LocalMultiplayerAgent
             VmConfiguration vmConfiguration,
             MultiLogger logger,
             ISessionHostRunnerFactory sessionHostRunnerFactory,
-            BasicAssetExtractor basicAssetExtractor)
+            BasicAssetExtractor basicAssetExtractor = null)
         {
             _sessionHostRunnerFactory = sessionHostRunnerFactory;
             _systemOperations = systemOperations;
             _logger = logger;
             _vmConfiguration = vmConfiguration;
-            _basicAssetExtractor = basicAssetExtractor;
+            _basicAssetExtractor = new BasicAssetExtractor(_systemOperations, _logger);
         }
 
         public async Task CreateAndStartContainerWaitForExit(SessionHostsStartInfo startParameters)
--- a/LocalMultiplayerAgent/MultiplayerServerManager.cs
+++ b/LocalMultiplayerAgent/MultiplayerServerManager.cs
@@ -54,8 +54,6 @@ namespace Microsoft.Azure.Gaming.LocalMultiplayerAgent
                 }
             }
 
-            BasicAssetExtractor.Instance = new BasicAssetExtractor(_systemOperations, _logger, Globals.ZipExtension, Globals.TarExtension);
-
             DownloadAndExtractAllAssets(startParameters);
             DownloadGameCertificates(startParameters);
 
@@ -153,7 +151,7 @@ namespace Microsoft.Azure.Gaming.LocalMultiplayerAgent
         {
             string assetFileName = _vmConfiguration.GetAssetDownloadFileName(assetDetail.assetPath);
 
-            BasicAssetExtractor.Instance.ExtractAssets(assetFileName,
+            _basicAssetExtractor.ExtractAssets(assetFileName,
                 _vmConfiguration.GetAssetExtractionFolderPathForSessionHost(0, assetDetail.assetNumber));
         }
     }
--- a/LocalMultiplayerAgent/MultiplayerServerManager.cs
+++ b/LocalMultiplayerAgent/MultiplayerServerManager.cs
@@ -14,6 +14,7 @@ namespace Microsoft.Azure.Gaming.LocalMultiplayerAgent
     using System.Threading.Tasks;
     using AgentInterfaces;
     using Config;
+    using Microsoft.Azure.Gaming.VmAgent.Core.Dependencies;
     using VmAgent.Core;
     using VmAgent.Core.Interfaces;
     using VmAgent.Model;
@@ -24,17 +25,20 @@ namespace Microsoft.Azure.Gaming.LocalMultiplayerAgent
         private readonly ISessionHostRunnerFactory _sessionHostRunnerFactory;
         private readonly MultiLogger _logger;
         private readonly VmConfiguration _vmConfiguration;
+        private readonly BasicAssetExtractor _basicAssetExtractor;
 
         public MultiplayerServerManager(
             ISystemOperations systemOperations,
             VmConfiguration vmConfiguration,
             MultiLogger logger,
-            ISessionHostRunnerFactory sessionHostRunnerFactory)
+            ISessionHostRunnerFactory sessionHostRunnerFactory,
+            BasicAssetExtractor basicAssetExtractor)
         {
             _sessionHostRunnerFactory = sessionHostRunnerFactory;
             _systemOperations = systemOperations;
             _logger = logger;
             _vmConfiguration = vmConfiguration;
+            _basicAssetExtractor = basicAssetExtractor;
         }
 
         public async Task CreateAndStartContainerWaitForExit(SessionHostsStartInfo startParameters)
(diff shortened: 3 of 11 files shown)


LocalMultiplayerAgent/MultiplayerServerManager.cs
```diff
@@ -163,6 +164,72 @@ private void ExtractAssets(string assetFileName, string targetFolder)
...
             }
+            else
+            {
+                _systemOperations.CreateDirectory(targetFolder);
+
+                ProcessStartInfo processStartInfo = Path.GetExtension(assetFileName).ToLowerInvariant() == Globals.ZipExtension
+                   ? GetProcessStartInfoForZip(assetFileName, targetFolder)
+                   : GetProcessStartInfoForTarOrGZip(assetFileName, targetFolder);
+
+                _logger.LogInformation($"Starting asset extraction with command arguments: {processStartInfo.Arguments}");
```
pedrofv, line 175: >LogInformation [](http://example.com/codeflow?start=24&length=14)

Should this be Log Debug or Verbose?

emmayspark: Thanks Pedro! I didn't put a lot of attention to Logging Levels and just learned there are different logging levels for different purposes. I think I can replace this to LogVerbose.

LocalMultiplayerAgent/Config/MultiplayerSettingsValidator.cs
```diff
@@ -35,11 +35,11 @@ public bool IsValid()
                 throw new Exception("OutputFolder or TitleId not found. Call SetDefaultsIfNotSpecified() before this method");
             }
 
-            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
-            {
-                Console.WriteLine("Running LocalMultiplayerAgent on Linux is not supported yet.");
-                return false;
-            }
+            //if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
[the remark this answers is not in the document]
emmayspark: I added validation code and removed commented part :)

LocalMultiplayerAgent/MultiplayerServerManager.cs
ravarnamsft: For local multiplayer agent, I don't think it makes much difference. You can keep it as is. For the actual agent, info means it will be sent to Geneva (which should be fine for asset extraction since it only happens once on a VM). Verbose are only written to local files on the VM and never sent to GEneva.

ravarnamsft, line 175: this code is copy pasted from VmAgent right? We should just move it to a separate AssetExtraction class and use the same one in both places

ravarnamsft (changes requested):
:clock1:

LocalMultiplayerAgent/MultiplayerServerManager.cs
emmayspark: Thanks for the feedback Rakesh. Does it mean we just need to create separate file (for example) **AssetProcessStartInfo.cs** and put that under VmAgent.Core? Do you think there is a scenario in the future LocalMultiplayerAgent will use asset extraction code (defined in VmAgent) like AssetExtraction.cs, AssetExtractorFactory.cs, ZipExtracion.cs and etc... These files currently used when VmAgent download asset as stream but not sure if I need to move those files too.

ravarnamsft: YEah I would call it Basic(or Simple)AssetExtractor.cs and put the ExtractAsset() method along with the private methods for GetProcessStartInfo in VmAgent.Core. We don't need the AssetExtractorFactory and other ones here yet. They will moved if and when we open source the rest of it. They are specific for streaming that's not used here.

emmayspark: Makes sense. Thanks :)

LocalMultiplayerAgent/Globals.cs
```diff
@@ -30,6 +30,11 @@ public static class Globals
 
         public static GameServerEnvironment GameServerEnvironment { get; set; }
 
+        public const string ZipExtension = ".zip";
```
ravarnamsft, line 33: We should move this to the Core library. So that the Asset extractor has access to these directly, without needing to be passed in (otherwise we will have to pass in again from actual VnAgent as well). Especially these are static strings that don't change so all classes should have access to them.

VmAgent.Core/Dependencies/BasicAssetExtractor.cs
```diff
@@ -0,0 +1,109 @@
...
+
+namespace Microsoft.Azure.Gaming.VmAgent.Core.Dependencies
+{
+    public class BasicAssetExtractor
+    {
+        public static BasicAssetExtractor Instance = new BasicAssetExtractor();
+
+        private readonly ISystemOperations _systemOperations;
+        private readonly MultiLogger _logger;
+        private string _zipExtension;
```
emmayspark, line 17: Do we need to keep file extension parameters in Globals? Since we moved these methods under VmAgent.Core from VmAgent, need to figure it out how to access Global parameters.

ravarnamsft: I just added a comment about it :)

VmAgent.Core/Dependencies/BasicAssetExtractor.cs
```diff
@@ -0,0 +1,109 @@
...
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Runtime.InteropServices;
+using System.Text;
+
+namespace Microsoft.Azure.Gaming.VmAgent.Core.Dependencies
+{
+    public class BasicAssetExtractor
```
ravarnamsft, line 11: Consider adding tests for this class

LocalMultiplayerAgent/MultiplayerServerManager.cs
```diff
@@ -50,6 +54,8 @@ public async Task CreateAndStartContainerWaitForExit(SessionHostsStartInfo start
                 }
             }
 
+            BasicAssetExtractor.Instance = new BasicAssetExtractor(_systemOperations, _logger, Globals.ZipExtension, Globals.TarExtension);
```
ravarnamsft, line 57: >BasicAssetExtractor [](http://example.com/codeflow?start=47&length=19)

this should be initialized during startup and not here.

LocalMultiplayerAgent/MultiplayerServerManager.cs
```diff
@@ -147,22 +152,9 @@ private IList<PortMapping> GetPortMappings(SessionHostsStartInfo request, int se
...
-            if (_systemOperations.IsOSPlatform(OSPlatform.Windows))
-            {
-                if (Directory.Exists(targetFolder))
-                {
-                    Directory.Delete(targetFolder, true);
-                }
-
-                _systemOperations.ExtractToDirectory(assetFileName, targetFolder);
-            }
+            BasicAssetExtractor.Instance.ExtractAssets(assetFileName,
```
ravarnamsft, line 156: this should be passed in and we should use _basicAssetExtractor

LocalMultiplayerAgent/Globals.cs
emmayspark: Thanks. But there are some other parts still need global file extension variables. For example in GameResourceManger.cs, there is SupportedAssetArchiveFormats variable ValidateAssetType() method still need to access it (VmAgent). what about creating a static class has these 3 file extension variables? Or should we keep that in VmAgent Side?

ravarnamsft: We should move it to the Core library. We can have a static class called Constants.cs and put those in there.

LocalMultiplayerAgent/Startup.cs
```diff
@@ -36,10 +39,13 @@ public void ConfigureServices(IServiceCollection services)
                     options.SerializerSettings.NullValueHandling = NullValueHandling.Ignore;
                     options.SerializerSettings.Converters.Add(new StringEnumConverter());
                 });
+            
+            BasicAssetExtractor.Instance = new BasicAssetExtractor(SystemOperations.Default, Globals.MultiLogger);
```
emmayspark, line 43: Is it preferable to add BasicAssetExtractor as singleton service here as the way Startup.cs in VmAgent implemented?

LocalMultiplayerAgent/MultiplayerServerManager.cs
emmayspark: updated !

VmAgent.Core/Dependencies/BasicAssetExtractor.cs
```diff
@@ -0,0 +1,105 @@
...
+    {
+        public static BasicAssetExtractor Instance = new BasicAssetExtractor();
+
+        private readonly ISystemOperations _systemOperations;
+        private readonly MultiLogger _logger;
+
+        public BasicAssetExtractor(ISystemOperations systemOperations = null, MultiLogger logger = null)
```
pedrofv, line 18: >= null [](http://example.com/codeflow?start=70&length=6)

Should this constructor use the defaults:

new BasicAssetExtractor(SystemOperations.Default, Globals.MultiLogger)

Otherwise the problem I see is that just doing 'new BasicAssetExtractor()' with null defaults does not give you an instance that actually works (I think it will throw a NullRef error if you try calling 'ExtractAssets()').

emmayspark: Thanks Pedro, I put this code **BasicAssetExtractor.Instance = new BasicAssetExtractor(SystemOperations.Default, Globals.MultiLogger)** in startup.cs to instantiate it. you think I need to update it? I also added the comment for it above ! :)

LocalMultiplayerAgent/Startup.cs
pedrofv: I think this is fine, especially since we're inside the method 'ConfigureServices' so this shows the extractor being configured :)

Nit: Personally I think we may want to avoid using a singleton since a potential issue with using the singleton is there is no way to know if the singleton was initialized properly for the OS we want and it could lead to an unexpected error.

VmAgent.Core.UnitTests/ExceptionAssert.cs
```diff
@@ -0,0 +1,36 @@
+﻿namespace Microsoft.Azure.Gaming.VmAgent.Core.UnitTests
```
emmayspark, line 1: This code is from VmAgent. Once I get confirmed that unit testes I implemented needs this code, I will move this file from VmAgent to VmAgent core so they can share this file.

VmAgent.Core/Dependencies/BasicAssetExtractor.cs
```diff
@@ -0,0 +1,109 @@
...
+            _systemOperations = systemOperations;
+            _logger = logger;
+            _zipExtension = zipExtension;
+            _tarExtension = tarExtension;
+        }
+
+        public void ExtractAssets(string assetFileName, string targetFolder)
+        {
+            // If the OS is windows use the native .NET zip extraction (we only support .zip for Windows).
+            if (_systemOperations.IsOSPlatform(OSPlatform.Windows))
```
emmayspark, line 29: I'm not sure if we need to add the code here that checks if file asset name is ending with supported format in VmAgent (.tar, .tar.gz, .zip), because we already have validation code on front side (Control Plane, and Web Portal)

VmAgent.Core/Dependencies/BasicAssetExtractor.cs
```diff
@@ -0,0 +1,108 @@
...
+using System.IO;
+using System.Runtime.InteropServices;
+using System.Text;
+using VmAgent.Core.Interfaces;
+
+namespace Microsoft.Azure.Gaming.VmAgent.Core.Dependencies
+{
+    public class BasicAssetExtractor: IBasicAssetExtractor
+    {
+        public static BasicAssetExtractor Instance = new BasicAssetExtractor();
```
ravarnamsft, line 15: Singletons should not have a public setter, which this has. this should be defined as :
public static BasicAssetExtractor Instance {get;} = new BasicAssetExtractor();

emmayspark: Thanks for the feedback! I'm not sure if we can access globals from BasicAssetExtractors considering Globals.cs is under LocalMultiplayerAgent and BasicAssetExtractos is under VmAgent.Core.

ravarnamsft: Perhaps we can remove the logger for now and log that line while calling the ExtractAssets method. IT is not the greatest solution, but gives a way out for now. As for why not a static class, it needs input from outside to set up the loggers within it. An in general static classes have a very small set of specific use cases that we should use them for.

emmayspark: What about this approach? Would it be not recommended?

public static BasicAssetExtractor Instance { get; } = new BasicAssetExtractor();

        private readonly ISystemOperations _systemOperations;
        private readonly MultiLogger _logger;

        public BasicAssetExtractor(ISystemOperations systemOperations = null, MultiLogger logger = null)
        {
            _systemOperations = systemOperations ?? SystemOperations.Default;
            _logger = logger;
        }

ravarnamsft: You have to still instantiate Instance outside of this class right? like you are doing it in the Program.cs. IF that's the case, we should remove the Instance completely. You can then just create a new instance wherever you want rather than trying to use a singleton Instance. It is perhaps better that way.

LocalMultiplayerAgent/Config/MultiplayerSettingsValidator.cs
```diff
@@ -35,9 +35,9 @@ public bool IsValid()
                 throw new Exception("OutputFolder or TitleId not found. Call SetDefaultsIfNotSpecified() before this method");
             }
 
-            if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux) && _settings.RunContainer)
```
dgkanatsios, line 38: this will give "true" if the OS is MacOS. Do we want to support that at this time? This "may just work" ([https://github.com/PlayFab/MpsAgent/pull/30](https://github.com/PlayFab/MpsAgent/pull/30)) but not sure till we test it out

VmAgent.Core/Dependencies/BasicAssetExtractor.cs
emmayspark: I think if we remove the logger and make sure SystemOperations.Default is initialized before the instance is used, we don't need to instantiate outside.  Another way as you suggested, I can also remove instance completely and create a new instance like code below. This case we can keep loggers. If I missed the point, please correct me! 

public MultiplayerServerManager(
            ISystemOperations systemOperations,
            VmConfiguration vmConfiguration,
            MultiLogger logger,
            ISessionHostRunnerFactory sessionHostRunnerFactory,
            BasicAssetExtractor basicAssetExtractor = null )
        {
            _sessionHostRunnerFactory = sessionHostRunnerFactory;
            _systemOperations = systemOperations;
            _logger = logger;
            _vmConfiguration = vmConfiguration;
            _basicAssetExtractor = new basicAssetExtractor(_systemOperations, _logger);
        }

LocalMultiplayerAgent/Config/MultiplayerSettingsValidator.cs
emmayspark: Thanks Dimitris, I will test it out and see if we need to consider MacOS case!

VmAgent.Core/Dependencies/BasicAssetExtractor.cs
ravarnamsft: The code above seems fine to me. You can keep the logger and remove the Instance

VmAgent.Core.UnitTests/BasicAssetExtractorTest.cs
```diff
@@ -29,62 +30,69 @@ public class BasicAssetExtractorTest
...
+
             _mockSystemOperations.Setup(x => x.RunProcessWithStdCapture(It.IsAny<Process>())).Returns((1, "stdOut", "stdErr"));
 
-            ExceptionAssert.Throws<Exception>(() => _basicAssetExtractor.ExtractAssets(_TarAssetFileName, _targetFolder));
+            ExceptionAssert.Throws<AssetExtractionFailedException>(() => _basicAssetExtractor.ExtractAssets(asset, _targetFolder));
         }
 
         [TestMethod, TestCategory("BVT")]
-        public void VerifyProcessStartInfoForTar()
+        [DataRow("")]
```
emmayspark, line 58: I'm not sure if adding multiple DataRows is the best way. I checked how to use DynamicData to find a better way to provide the test dataset to each unit tests but seems it's unnecessary. Happy to hear any feedback!

VmAgent.Core/Dependencies/BasicAssetExtractor.cs
emmayspark: updated

LocalMultiplayerAgent/Config/MultiplayerSettingsValidator.cs
emmayspark: I will add **if RuntimeInformation.IsOSPlatform(OSPlatform.OSX) them throw some exception with message** because we still don't know if we are supping OS or not. any thought? :)

VmAgent.Core/Dependencies/BasicAssetExtractor.cs
ravarnamsft: The agent code was written before/separately from the front end. So we could theoretically remove the checks, but no harm in keeping it either.

ravarnamsft (approved):
:shipit:

LocalMultiplayerAgent/Config/MultiplayerSettingsValidator.cs
dgkanatsios: that's great - I would even add "we would be happy to accept PRs to make it work" :)

LocalMultiplayerAgent/Config/MultiplayerSettingsValidator.cs
```diff
@@ -35,9 +35,15 @@ public bool IsValid()
...
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
             {
-                Console.WriteLine("Running LocalMultiplayerAgent on Linux is not supported yet.");
+                Console.WriteLine("Running LocalMultiplayerAgent is not yet supported on MacOS.");
+                return false;
+            }
+
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux) && _settings.RunContainer)
+            {
+                Console.WriteLine("Running LocalMultiplayerAgent as container mode is not yet supported on Linux. Please set RunContainer to false in MultiplayerSettings.json");
```
dgkanatsios, line 46: (not for this PR but in general...) this is kind of cumbersome, supporting running Linux Containers under Windows and not under Linux. We should fix that.

VmAgent.Core/Dependencies/BasicAssetExtractor.cs
```diff
@@ -0,0 +1,123 @@
...
+                }
+
+                if (Directory.Exists(targetFolder))
+                {
+                    Directory.Delete(targetFolder, true);
+                }
+
+                _systemOperations.ExtractToDirectory(assetFileName, targetFolder);
+            }
+            else
```
dgkanatsios, line 48: add a comment that we're running as Linux here so it's visible - since we blocked running on MacOS

dgkanatsios (approved):
:shipit:

PR #87, opened 2022-01-20
More CR comments from PR#85

pgilmorepf:


raymondchams (approved):

LocalMultiplayerAgent/SettingsJsonGenerator/scripts.js
```diff
@@ -2,6 +2,27 @@
 let BuildGuId = crypto.randomUUID();
 let SessionGuId = crypto.randomUUID()
 
+// These values match an enum for RunMode in the HTML file
+let RUN_MODE_WIN_PROCESS = "WinProcess";
```
nassosterz-ms, line 6: Maybe have all these be const?

raymondchams: I think 'let' is const in js.

dgkanatsios (approved):

emmayspark:
Are we going to remove the current Json configuration permanently or supporting both?

LocalMultiplayerAgent/SettingsJsonGenerator/scripts.js
pgilmorepf: No. It's only about scoping.
https://stackoverflow.com/questions/762011/whats-the-difference-between-using-let-and-var
Basically let is more precisely similar to the way ALL C++/C#/etc languages work, where var is abusing the underlying language implementation in ways that "shouldn't technically work, but do".

pgilmorepf:
E: This doesn't replace it. It's a generator for it. The Json configuration stays, but this webpage can generate it from a simple gui of options, which also has range-checking, validation, and other helpful inputs to ensure you have input the correct values.

LocalMultiplayerAgent/SettingsJsonGenerator/scripts.js
nassosterz-ms: I am not familiar with JS, but what about this? https://www.tutorialspoint.com/const-vs-let-in-javascript

pgilmorepf: I was wrong! Fixed.

PR #90, opened 2022-01-25
Improvements to LocalMultiplayerAgent MultiplayerSettings.json Generator

jordan-playfab:
It looks nicer now. I can keep going if you want more changes, but the basic layout was fine.

- Added semantic form HTML
- Styled page
- Converted JavaScript to TypeScript
- Added [Tippy](https://atomiks.github.io/tippyjs/) tooltips (add `data-tippy-content` to elements to enable)

Screenshot 2022-01-25 150314


dgkanatsios:
What does the user need to do to run the file now? Previously they would just have to download the .html and the .js.

jordan-playfab:
Nothing has changed for the user. There are more files in the root folder, for which I apologize, but the process is the same: open index.html in a web browser.

dgkanatsios:
thank you. Can they do it locally (i.e. if the .html page is saved in their Desktop)?

jordan-playfab:
They can view it locally, same as before. We can and should host it somewhere, but that's not necessary.

dgkanatsios (approved):
great work, thanks!

pgilmorepf (approved):
I like this so much!

PR #92, opened 2022-02-03
Test to ensure enum order doesn't change when a new VmState is added

sarahwangsw:
this is to add a test for the comment in this PR: https://dev.azure.com/playfab/PlayFab/_git/Thunderhead/pullrequest/11451
"Although unlikely to happen, it might be worth adding a test to ensure the order of enums does not change. In general, new enum values are added at the bottom of the list. Sometimes, people do miss that and add it in between which could then break this assumption"

we rely on this enum to determine the order healthyVmsEligibleToRelease . This test is added to ensure enum order doesn't change.

### DIFF
--- a/AgentInterfaces.Tests/VmAgentInfoTests.cs
+++ b/AgentInterfaces.Tests/VmAgentInfoTests.cs
@@ -64,27 +64,30 @@ namespace AgentInterfaces.Tests
         /// </summary>
         [TestMethod]
         [TestCategory("BVT")]
-        public void VmStateEnumOrder()
+        public void KnowVmStateOrder()
         {
-            var vmStatesEnumIntCast = new Dictionary<VmState, int>
+           //we need to keep the order vm state because we use this order to determine healthy vms to release
+            var knownVmStateOrder = new Dictionary<string, int>
             {
-                {VmState.Unknown, 0},
-                {VmState.Unassigned, 1},
-                {VmState.Assigned, 2},
-                {VmState.Propping, 3},
-                {VmState.ProppingFailed, 4},
-                {VmState.ProppingCompleted, 5},
-                {VmState.ServerStartFailed, 6},
-                {VmState.StartServersFailed, 7},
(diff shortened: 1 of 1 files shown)


AgentInterfaces.Tests/VmAgentInfoTests.cs
```diff
@@ -58,5 +58,34 @@
...
+                {VmState.Propping, 3},
+                {VmState.ProppingFailed, 4},
+                {VmState.ProppingCompleted, 5},
+                {VmState.ServerStartFailed, 6},
+                {VmState.StartServersFailed, 7},
+                {VmState.PartiallyRunning, 8},
+                {VmState.Running, 9},
+                {VmState.PendingResourceCleanup, 10},
+                {VmState.ServersRemoved, 12},
+                {VmState.TooManyServerRestarts, 13}
```
ravarnamsft, line 83: As dscussed offline, the idea of the test is to ensure new values are only added at the end. So we can keep  a list of the names of known states (which is the current set of values and then check that there are no new ones). The error message can tell the person to ensure adding it at the end

sarahwangsw: on a second thought, I think we want to release vms that are furthest from being ready to running. so this order can potentially change, but as long as it keeps the order of how a vm transitions, it's fine? if we get a new vm state, such as "almost running state", it shouldn't be added to the end of the list but right before running state. how do you think?

ravarnamsft: new enum values should not be added in the middle, that could potentially break values stored externally and parsing a previous value. For the case you mentioned, we should then not consider the enums as an order of the state transitions and instead categorize them separately or maintain a separate state transition map

sarahwangsw: sounds good! I think the test here can address new states are added at the end of the list.

PR #93, opened 2022-02-11
Add a couple of settings for improving game availability.

ravarnamsft:
- A setting to pass the maintenance info to games so that they can potentially migrate the game out of the VM.
- A setting to determine whether we should clean up game assets that already exist on the VM or reuse them. 
- Removed a noisy log.

### DIFF
--- a/VmAgent.Core/Interfaces/SystemOperations.cs
+++ b/VmAgent.Core/Interfaces/SystemOperations.cs
@@ -318,10 +318,6 @@ namespace Microsoft.Azure.Gaming.VmAgent.Core.Interfaces
                     }
                 }
             }
-            else
-            {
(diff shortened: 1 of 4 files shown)


VmAgent.Core/Interfaces/SystemOperations.cs
```diff
@@ -318,10 +318,6 @@ public void SetUnixOwnerIfNeeded(string path, bool applyToAllContents = false)
                     }
                 }
             }
-            else
-            {
-                _logger?.LogVerbose("Unix file setting not needed");
```
dgkanatsios, line 323: many thank for removing that one!

dgkanatsios (approved):

PR #97, opened 2022-02-13
Add generator files to the Build output

dgkanatsios:
This PR adds the JSON settings generator files to the output. FYI @pgilmorepf @jordan-playfab 
Also setting a checkbox that configures required values set to checked as default

### DIFF
--- a/LocalMultiplayerAgent/LocalMultiplayerAgent.csproj
+++ b/LocalMultiplayerAgent/LocalMultiplayerAgent.csproj
@@ -52,6 +52,18 @@
     <Content Include="SetupLinuxContainersOnWindows.ps1">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </Content>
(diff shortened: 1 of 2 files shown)


emmayspark (approved):

pgilmorepf:
Addin @jordan-playfab, as I recall there being more files in his updated version, and I'm not sure how it interacts with VS (his readme instructions don't reflect any interaction with VS).

dgkanatsios:
FWIW, adding the html/js/css files in the csproj in this PR didn’t work in getting them as output during publishing <sigh>
something broke in dotnet publish command after .NET Core 3.1

pgilmorepf:
That's kinda what I suspected. That webpage just isn't designed to be integrated into that sln.

dgkanatsios:
Webpage with the files included (the two CSS and the one js) works just fine with “dotnet build” command, the problem is with “dotnet publish”.

PR #103, opened 2022-03-08
expose GetFiles method in SystemOperations

brianwp3000:
I need this method for process-based crash dumps in VmAgent


pedrofv (approved):
:shipit:

PR #112, opened 2022-06-11
Added logging for game not found bug

yegarci:
The bug ADO Task in question:

- [[VmAgent] Process-based servers don't capture logs on bad startup command](https://microsoft.visualstudio.com/Xbox/_sprints/taskboard/AGS%20MP%20Hosting/Xbox/22-4W24%20(Ends%206-10)/2W24%20(Ends%206-10)?workitem=33466233)

In summary, if a user is testing their game.exe but the file is not found, there is no logging done currently to make the user aware of the issue. This PR focuses on Windows-based processes.

I've added functionality to a separate CollectLogs() method within ProcessRunner.cs that is only called if the process failed to start. If the game starts successfully, no log file is created in the destination directory. If the game fails, a file with the exception message is created in the destination directory.

To test:
1. You can use the MultiplayerSettings.json used in MpsSamples. Edit "OutputFolder" path with a local folder path of your choosing. Edit "StartGameCommand" to have a game.exe that does not exist (i.e. nonExistent.exe).
2. Delete LocalMultiplayerAgentPublishFolder.
3. Choose LocalMultiplayerAgent as your Startup Project. 
4. Create breakpoint at try{} catch{} blocks for _processWrapper.Start(...) on line 65 and 70, respectively, within ProcessRunner.cs.
5. Select LocalMultiplayerAgent as Build, and begin Debug.
6. Check file called PF_ConsoleLogs.txt was created in destination directory with helpful log that tells you "The system cannot find the file specified.".

Please let me know if you have any recommendations or better courses of actions. Just trying to get my first PR in and understanding this new environment :)

### DIFF
--- a/VmAgent.Core/Interfaces/DockerContainerEngine.cs
+++ b/VmAgent.Core/Interfaces/DockerContainerEngine.cs
@@ -37,12 +37,6 @@ namespace Microsoft.Azure.Gaming.VmAgent.ContainerEngines
         private readonly int _maxRetryAttempts = 30;
 
         private readonly bool _shouldPublicPortMatchGamePort;
-
-        /// <summary>
-        /// The name of the file where the console logs for the server are captured.
-        /// </summary>
-        private const string ConsoleLogCaptureFileName = "PF_ConsoleLogs.txt";
-
         /// <summary>
         /// Lazily instantiates the Docker client.
         /// </summary>
@@ -157,7 +151,7 @@ namespace Microsoft.Azure.Gaming.VmAgent.ContainerEngines
             {
                 hostConfig.CapAdd.Add("SYS_PTRACE");
             }
-            
+
 
             if(hostConfig.LogConfig == null)
             {
@@ -197,7 +191,7 @@ namespace Microsoft.Azure.Gaming.VmAgent.ContainerEngines
                 MetricConstants.ContainerStartTime);
             _logger.LogInformation($"Container {containerId} start completed.");
         }
-        
+
         private class LogReporter : IProgress<JSONMessage>
         {
             private readonly ConcurrentDictionary<string, LayerProgress> _layerProgresses = new ConcurrentDictionary<string, LayerProgress>();
@@ -360,7 +354,7 @@ namespace Microsoft.Azure.Gaming.VmAgent.ContainerEngines
             SessionHostsStartInfo sessionHostStartInfo = gameResourceDetails.SessionHostsStartInfo;
             ContainerImageDetails imageDetails = sessionHostStartInfo.ImageDetails;
             string imageName = $"{imageDetails.ImageName}:{imageDetails.ImageTag ?? "latest"}";
-            
+
             // Support running local images with no explicit registry in the name.
             if (!string.IsNullOrEmpty(imageDetails.Registry))
             {
@@ -409,14 +403,14 @@ namespace Microsoft.Azure.Gaming.VmAgent.ContainerEngines
             {
                 await StartContainer(dockerId);
                 _logger.LogInformation($"Started container {dockerId}, with assignmentId {sessionHostStartInfo.AssignmentId}, instance number {instanceNumber}, and logFolderId {logFolderId}");
-            } 
+            }
             catch (Exception exception)
             {
                 _logger.LogException($"Failed to start container based host with instance number {instanceNumber}", exception);
                 sessionHostManager.RemoveSessionHost(dockerId);
                 sessionHost = null;
             }
-           
+
 
             return sessionHost;
         }
@@ -586,7 +580,7 @@ namespace Microsoft.Azure.Gaming.VmAgent.ContainerEngines
                     throw new ApplicationException("CreateImageAsync is completed but the image doesn't exist");
                 }
             });
-            
+
             _logger.LogEvent(MetricConstants.PullImage, null, new Dictionary<string, double>
                 {
                     { MetricConstants.DownloadDurationInMilliseconds, logReporter.DownloadSummary?.DurationInMilliseconds ?? 0d },
@@ -614,7 +608,7 @@ namespace Microsoft.Azure.Gaming.VmAgent.ContainerEngines
                 if (_systemOperations.IsOSPlatform(OSPlatform.Windows))
                 {
                     var windowsStartCommand = new List<string>();
-                    
+
                     if (request.WindowsCrashDumpConfiguration?.IsEnabled == true)
                     {
                         // set crash dump registry keys on startup
@@ -664,6 +658,11 @@ namespace Microsoft.Azure.Gaming.VmAgent.ContainerEngines
                 }
             }
             return null;
+        }
+
+        public override Task CreateExceptionLogs(string logsFolder, string exceptionMessage)
+        {
+            throw new NotSupportedException();
         }
     }
 }
--- a/VmAgent.Core/Interfaces/DockerContainerEngine.cs
+++ b/VmAgent.Core/Interfaces/DockerContainerEngine.cs
@@ -660,9 +660,9 @@ namespace Microsoft.Azure.Gaming.VmAgent.ContainerEngines
             return null;
         }
 
-        public override Task CreateExceptionLogs(string logsFolder, string exceptionMessage)
+        public override Task CreateStartGameExceptionLogs(string logsFolder, string exceptionMessage)
         {
-            throw new NotSupportedException();
+            return Task.CompletedTask;
         }
     }
 }
--- a/VmAgent.Core/Interfaces/IProcessWrapper.cs
+++ b/VmAgent.Core/Interfaces/IProcessWrapper.cs
@@ -5,12 +5,17 @@ namespace Microsoft.Azure.Gaming.VmAgent.Core.Interfaces
 {
     using System.Collections.Generic;
     using System.Diagnostics;
+    using System.IO;
 
     public interface IProcessWrapper
     {
         void Kill(int id);
 
         int Start(ProcessStartInfo startInfo);
+
+        void StartErrorStreamReader(Process process);
+
+        StreamReader StandardErrorReader { get; }
 
         IEnumerable<int> List();
 
--- a/VmAgent.Core/Interfaces/IProcessWrapper.cs
+++ b/VmAgent.Core/Interfaces/IProcessWrapper.cs
@@ -13,10 +13,6 @@ namespace Microsoft.Azure.Gaming.VmAgent.Core.Interfaces
 
         int Start(ProcessStartInfo startInfo);
 
-        void StartErrorStreamReader(Process process);
-
-        StreamReader StandardErrorReader { get; }
-
         IEnumerable<int> List();
 
         void WaitForProcessExit(int id);
--- a/VmAgent.Core/Interfaces/IProcessWrapper.cs
+++ b/VmAgent.Core/Interfaces/IProcessWrapper.cs
@@ -5,7 +5,6 @@ namespace Microsoft.Azure.Gaming.VmAgent.Core.Interfaces
 {
     using System.Collections.Generic;
     using System.Diagnostics;
-    using System.IO;
 
     public interface IProcessWrapper
     {
--- a/VmAgent.Core/Interfaces/ISessionHostRunner.cs
+++ b/VmAgent.Core/Interfaces/ISessionHostRunner.cs
@@ -25,5 +25,7 @@ namespace Microsoft.Azure.Gaming.VmAgent.Core.Interfaces
         Task WaitOnServerExit(string containerId);
 
         Task CollectLogs(string id, string logsFolder, ISessionHostManager sessionHostManager);
+
+        Task CreateExceptionLogs(string logFolder, string exceptionMessage);
     }
 }
--- a/VmAgent.Core/Interfaces/ISessionHostRunner.cs
+++ b/VmAgent.Core/Interfaces/ISessionHostRunner.cs
@@ -26,6 +26,6 @@ namespace Microsoft.Azure.Gaming.VmAgent.Core.Interfaces
 
         Task CollectLogs(string id, string logsFolder, ISessionHostManager sessionHostManager);
 
-        Task CreateExceptionLogs(string logFolder, string exceptionMessage);
+        Task CreateStartGameExceptionLogs(string logFolder, string exceptionMessage);
     }
 }
--- a/VmAgent.Core/Interfaces/ProcessRunner.cs
+++ b/VmAgent.Core/Interfaces/ProcessRunner.cs
@@ -24,7 +24,7 @@ namespace Microsoft.Azure.Gaming.VmAgent.Core.Interfaces
             MultiLogger logger,
             ISystemOperations systemOperations,
             IProcessWrapper processWrapper = null)
-            : base(vmConfiguration, logger, systemOperations)
+            : base (vmConfiguration, logger, systemOperations)
         {
             _processWrapper = processWrapper;
         }
@@ -74,7 +74,7 @@ namespace Microsoft.Azure.Gaming.VmAgent.Core.Interfaces
             return Task.FromResult(sessionHost);
         }
 
-        public override Task CollectLogs(string id, string logsFolder, ISessionHostManager sessionHostManager)
+        public override Task CollectLogs(string processId, string logsFolder, ISessionHostManager sessionHostManager)
         {
             // The game server is free to read the env variable for log folder and write all output to a file in that folder.
             // If required, we can add action handlers (see SystemOperations.RunProcess for example).
--- a/VmAgent.Core/Interfaces/ProcessRunner.cs
+++ b/VmAgent.Core/Interfaces/ProcessRunner.cs
@@ -8,6 +8,7 @@ namespace Microsoft.Azure.Gaming.VmAgent.Core.Interfaces
     using System.Diagnostics;
     using System.IO;
     using System.Linq;
+    using System.Runtime.InteropServices;
     using System.Threading.Tasks;
     using AgentInterfaces;
     using Microsoft.Azure.Gaming.VmAgent.ContainerEngines;
@@ -18,17 +19,12 @@ namespace Microsoft.Azure.Gaming.VmAgent.Core.Interfaces
     {
         private readonly IProcessWrapper _processWrapper;
 
-        /// <summary>
-        /// The name of the file where the console logs for the server are captured.
-        /// </summary>
-        private const string ConsoleLogCaptureFileName = "PF_ConsoleLogs.txt";
-
         public ProcessRunner(
             VmConfiguration vmConfiguration,
             MultiLogger logger,
             ISystemOperations systemOperations,
             IProcessWrapper processWrapper = null)
-            : base (vmConfiguration, logger, systemOperations)
+            : base(vmConfiguration, logger, systemOperations)
         {
             _processWrapper = processWrapper;
         }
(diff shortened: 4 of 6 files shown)


VmAgent.Core/Interfaces/ProcessRunner.cs
```diff
@@ -76,6 +84,30 @@ public override Task CollectLogs(string processId, string logsFolder, ISessionHo
...
+                                sw.Restart(); 
+                            }
+                            _systemOperations.FileAppendAllText(destinationFileName, _processWrapper.StandardErrorReader.ReadLine() + Environment.NewLine); // if .ReadToEnd, use this at the end
+                        }
+                    }
+                    _logger.LogVerbose($"Written logs for process {processId} to {destinationFileName}.");
+            }
+            catch (Exception ex)
+            {
+                _logger.LogException($"Failed to write error logs from Windows process: {processId}", ex);
```
dgkanatsios, line 109: this functionality is to be used from both Windows and Linux processes, right?

VmAgent.Core/Interfaces/ProcessRunner.cs
```diff
@@ -66,16 +72,53 @@ public override Task<SessionHostInfo> CreateAndStart(int instanceNumber, GameRes
...
+                    using (StreamWriter sw = File.AppendText(destinationFileName))
+                    {
+                        try
+                        {
+                            sw.WriteLine("{0} {1}", DateTime.Now.ToLongTimeString(), DateTime.Now.ToLongDateString());
+                            sw.WriteLine($"{exceptionMessage}");
+                        }
+                        catch (Exception ex)
+                        {
+                            sw.WriteLine(ex.ToString());
```
brianwp3000, line 108: Is this try-catch necessary? It looks like the only thing that could fail in it is "sw.WriteLine", which is exactly what we do in the Catch block so it's just going to fail again

VmAgent.Core/Interfaces/ProcessRunner.cs
```diff
@@ -66,16 +72,53 @@ public override Task<SessionHostInfo> CreateAndStart(int instanceNumber, GameRes
...
+        {
+            return Task.CompletedTask;
+        }
+
+        public override Task CollectLogs(string logsFolder, string exceptionMessage)
         {
             // The game server is free to read the env variable for log folder and write all output to a file in that folder.
             // If required, we can add action handlers (see SystemOperations.RunProcess for example).
             // However, keeping the file handle around can be tricky.
+            if (_systemOperations.IsOSPlatform(OSPlatform.Windows))
```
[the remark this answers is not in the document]
yegarci: The bug was for Windows processes

yegarci:
> More big picture question--so this solution captures any exception that occurs on process start and logs that exception in PF_Console_Logs.txt. Did you look into redirecting console output from the process? If it's possible to get that working, that would solve both this issue as well

I did look into it, but I couldn't keep the StreamWriter up and running

yegarci:
Apologies, this PR is pretty messy now because of all the extra commits and files from restores

dgkanatsios (commented):
Can we remove the files on the LocalMultiplayerAgent/ folder? We can add this folder to .gitignore

LocalMultiplayerAgent/LocalMultiplayerAgentPublishFolder/SettingsJsonGenerator/package-lock.json
```diff
@@ -0,0 +1,3468 @@
+{
```
dgkanatsios, line 1: Any chance we can rebase so these differences are not part of the commit?

VmAgent.Core/Interfaces/ProcessRunner.cs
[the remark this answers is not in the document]
brianwp3000: We have Linux Process-based builds now too though--if this solution works for them as well (I'm not immediately seeing any reason it wouldn't) i don't see a reason to limit it to just windows

VmAgent.Core/Interfaces/ProcessRunner.cs
```diff
@@ -66,16 +72,46 @@ public override Task<SessionHostInfo> CreateAndStart(int instanceNumber, GameRes
...
             return Task.FromResult(sessionHost);
         }
 
-        public override Task CollectLogs(string processId, string logsFolder, ISessionHostManager sessionHostManager)
+        public override Task CollectLogs(string id, string logsFolder, ISessionHostManager sessionHostManager)
+        {
+            return Task.CompletedTask;
+        }
+
+        public override Task CollectLogs(string logsFolder, string exceptionMessage)
```
brianwp3000, line 87: Could this method be called `CreateExceptionLog` or something? It's not really doing the same thing as the existing `CollectLogs` method, and if i'm not mistaken the existing method will still get called after this.

LocalMultiplayerAgent/LocalMultiplayerAgentPublishFolder/MultiplayerSettings.json
```diff
@@ -0,0 +1,55 @@
+{
```
brianwp3000, line 1: The entire LocalMultiplayerAgentPublishFolder should probably be excluded

VmAgent.Core/Interfaces/ISessionHostRunner.cs
```diff
@@ -25,5 +25,6 @@ public interface ISessionHostRunner
         Task WaitOnServerExit(string containerId);
 
         Task CollectLogs(string id, string logsFolder, ISessionHostManager sessionHostManager);
+        Task CollectLogs(string logFolder, string exceptionMessage);
```
pedrofv, line 28: >Task [](http://example.com/codeflow?start=8&length=4)

Nit: spacing

VmAgent.Core/Interfaces/DockerContainerEngine.cs
```diff
@@ -478,7 +478,7 @@
~                    // which the host Windows machine does not have "copy" access to, to get the logs with a FileCopy
~                    // this is only supposed to run on LocalMultiplayerAgent running on lcow
~                    StringBuilder sb = new StringBuilder();
~                    Stream logsStream = await _dockerClient.Containers.GetContainerLogsAsync(containerId,
~                        new ContainerLogsParameters() {ShowStdout = true, ShowStderr = true});
~                    using (StreamReader sr = new StreamReader(logsStream))
~                    {
```
pedrofv: >NotImplementedException [](http://example.com/codeflow?start=22&length=23)

We may want to throw 'NotSupportedException' instead in this case

VmAgent.Core/Interfaces/IProcessWrapper.cs
```diff
@@ -5,6 +5,7 @@ namespace Microsoft.Azure.Gaming.VmAgent.Core.Interfaces
 {
     using System.Collections.Generic;
     using System.Diagnostics;
+    using System.IO;
```
pedrofv, line 8: >System [](http://example.com/codeflow?start=10&length=6)

There are no other changes, is this needed here?

VmAgent.Core/Interfaces/ProcessWrapper.cs
```diff
@@ -6,11 +6,14 @@
     using System;
     using System.Collections.Generic;
     using System.Diagnostics;
+    using System.IO;
     using System.Linq;
     using System.Threading.Tasks;
 
     public class ProcessWrapper : IProcessWrapper
     {
+        StreamReader ErrorStreamReader = null;
```
pedrofv, line 15: >ErrorStreamReader [](http://example.com/codeflow?start=21&length=17)

If this is not needed, then we should remove

VmAgent.Core/Interfaces/ProcessRunner.cs
```diff
@@ -66,16 +72,46 @@ public override Task<SessionHostInfo> CreateAndStart(int instanceNumber, GameRes
...
+                CollectLogs(logFolderPathOnVm, exceptionMessage);
             }
 
             return Task.FromResult(sessionHost);
         }
 
-        public override Task CollectLogs(string processId, string logsFolder, ISessionHostManager sessionHostManager)
+        public override Task CollectLogs(string id, string logsFolder, ISessionHostManager sessionHostManager)
+        {
+            return Task.CompletedTask;
```
pedrofv, line 84: >return Task.CompletedTas [](http://example.com/codeflow?start=12&length=24)

Curious why we wouldn't return 'NotSupportedException' in this case, similar to what we do in DockerContainerEngine.cs for the other method?

VmAgent.Core/Interfaces/ProcessRunner.cs
```diff
@@ -66,16 +72,46 @@ public override Task<SessionHostInfo> CreateAndStart(int instanceNumber, GameRes
...
+            if (_systemOperations.IsOSPlatform(OSPlatform.Windows))
+            {
+                try
+                {
+                    _logger.LogVerbose("Collecting logs for failed start game process.");
+                    string destinationFileName = Path.Combine(logsFolder, ConsoleLogCaptureFileName);
+
+                    using (StreamWriter sw = File.AppendText(destinationFileName))
+                    {
+                        sw.WriteLine("{0} {1}", DateTime.Now.ToLongTimeString(), DateTime.Now.ToLongDateString());
```
pedrofv, line 101: >ToLongTimeString [](http://example.com/codeflow?start=61&length=16)

You can consolidate these to calls into a single one using:
DateTime.Now.ToString(" < enter format here > ")

You can see the available formats and examples here: 
[https://docs.microsoft.com/en-us/dotnet/standard/base-types/custom-date-and-time-format-strings](https://docs.microsoft.com/en-us/dotnet/standard/base-types/custom-date-and-time-format-strings)

VmAgent.Core/Interfaces/ProcessRunner.cs
```diff
@@ -66,16 +72,46 @@ public override Task<SessionHostInfo> CreateAndStart(int instanceNumber, GameRes
...
+                    _logger.LogVerbose($"Written logs for failed start game process to {destinationFileName}.");
+                }
+                catch (Exception ex)
+                {
+                    _logger.LogException($"Failed to write failed start game error logs for Windows process", ex);
+                }
+                return Task.CompletedTask;
+            }
+
+            _logger.LogVerbose("Logs not written to file for non-Windows operating systems.");
```
pedrofv, line 113: >Logs not written to file for non-Windows operating system [](http://example.com/codeflow?start=32&length=57)

We should confirm if we can support Linux as well.

Otherwise I would suggest we tweak the message to something like "Error logs from process execution are only available in Windows"  (or something similar)

VmAgent.Core/Interfaces/ProcessRunner.cs
yegarci: It was returning this before this PR, but I can change it to 'NotSupportedException' if that's preferred

VmAgent.Core/Interfaces/ProcessRunner.cs
yegarci: @dgkanatsios ?

dgkanatsios: The issue with Linux process game servers is that they are harder to test since they are not supported on LocalMultiplayerAgent at this time.

LocalMultiplayerAgent/SettingsJsonGenerator/index.js
```diff
@@ -1 +1 @@
.../Data/")&&(t="We recommended you choose a sub-folder of /Data"),document.getElementById("MountPathValidate").innerHTML=e,document.getElementById("MountPathWarning").innerHTML=t}()}const g=Array.prototype.slice.call(document.getElementsByTagName("input"),0),y=Array.prototype.slice.call(document.getElementsByTagName("select"),0);g.forEach((e=>{e.addEventListener("change",(()=>{const t=!i(e.getAttribute("min")),n=!i(e.getAttribute("max"));if(!isNaN(Number(e.value))&&(t||n)){let r=Number(e.value);if(t){const t=Number(e.getAttribute("min"));r<t&&(r=t)}if(n){const t=Number(e.getAttribute("max"));r>t&&(r=t)}e.value=r}v()}))})),y.forEach((e=>{e.addEventListener("change",(()=>{v()}))})),v(),(0,o.default)("[data-tippy-content]")}},t={};function n(r){var o=t[r];if(void 0!==o)return o.exports;var i=t[r]={exports:{}};return e[r].call(i.exports,i,i.exports,n),i.exports}n.d=(e,t)=>{for(var r in t)n.o(t,r)&&!n.o(e,r)&&Object.defineProperty(e,r,{enumerable:!0,get:t[r]})},n.o=(e,t)=>Object.prototype.hasOwnProperty.call(e,t),n.r=e=>{"undefined"!=typeof Symbol&&Symbol.toStringTag&&Object.defineProperty(e,Symbol.toStringTag,{value:"Module"}),Object.defineProperty(e,"__esModule",{value:!0})},n(607)})();
```
pedrofv, line 1: >(()=>{" [](http://example.com/codeflow?start=0&length=7)

I can't see any changes in this 'SettingsJsonGenerator' folder, should it be excluded from the PR?

VmAgent.Core/Interfaces/ProcessRunner.cs
pedrofv: I think we should be consistent in both scenarios, so let's change to return the exception if possible.

.gitignore
```diff
@@ -364,3 +364,5 @@ MigrationBackup/
 # Fody - auto-generated XML schema
 FodyWeavers.xsd
 
+C:\Users\yegarci\Source\Repos\MpsAgent\LocalMultiplayerAgent\*
```
emmayspark, line 367: we should remove this :)

VmAgent.Core/Interfaces/ProcessRunner.cs
```diff
@@ -18,6 +19,11 @@ public class ProcessRunner : BaseSessionHostRunner
     {
         private readonly IProcessWrapper _processWrapper;
 
+        /// <summary>
+        /// The name of the file where the console logs for the server are captured.
+        /// </summary>
+        private const string ConsoleLogCaptureFileName = "PF_ConsoleLogs.txt";
```
emmayspark, line 25: maybe we can move this to BaseSessionHostRunner because now this file name is defined in both ProcessRunner.cs and DockerContainerEngine.cs?

dgkanatsios:
>         public override Task CollectLogs(string processId, string logsFolder, ISessionHostManager sessionHostManager)

is this method called anywhere? If so, this will break existing callers since it now throws.

---
Refers to: VmAgent.Core/Interfaces/ProcessRunner.cs:77 in b222b25. [](commit_id = b222b2550f83e8ae8345fa6f5965f0d7755eb97b, deletion_comment = False)

VmAgent.Core/Interfaces/ProcessRunner.cs
```diff
@@ -76,6 +79,27 @@ public override Task CollectLogs(string processId, string logsFolder, ISessionHo
             // The game server is free to read the env variable for log folder and write all output to a file in that folder.
             // If required, we can add action handlers (see SystemOperations.RunProcess for example).
             // However, keeping the file handle around can be tricky.
+            throw new NotSupportedException();
+        }
+
+        public override Task CreateExceptionLogs(string logsFolder, string exceptionMessage)
```
dgkanatsios, line 85: This method is to be used only when the startup of the game server fails, correct? Maybe give it a more suitable naming then.

VmAgent.Core/Interfaces/ProcessRunner.cs
```diff
@@ -76,6 +79,27 @@ public override Task CollectLogs(string processId, string logsFolder, ISessionHo
...
+                using (StreamWriter sw = File.AppendText(destinationFileName))
+                {
+                    sw.WriteLine(DateTime.Now.ToString("HH:mm:ss D"));
+                    sw.WriteLine($"{exceptionMessage}");
+                }
+                _logger.LogVerbose($"Written logs for failed start game process to {destinationFileName}.");
+            }
+            catch (Exception ex)
+            {
+                _logger.LogException($"Failed to write failed start game error logs for Windows process", ex);
```
dgkanatsios, line 101: this method will be called on Linux process game servers as well (since there is no "if" for it), maybe remove the "Windows" from the logging message?

VmAgent.Core/Interfaces/DockerContainerEngine.cs
```diff
@@ -614,7 +608,7 @@
                 if (_systemOperations.IsOSPlatform(OSPlatform.Windows))
                 {
                     var windowsStartCommand = new List<string>();
-                    
+
                     if (request.WindowsCrashDumpConfiguration?.IsEnabled == true)
```
dgkanatsios, line 612: should we make it a no-op (return Task.CompletedTask) to avoid breaking anything in the future?

yegarci:
> > ```
> >     public override Task CollectLogs(string processId, string logsFolder, ISessionHostManager sessionHostManager)
> > ```
> 
> is this method called anywhere? If so, this will break existing callers since it now throws.
> 
> Refers to: VmAgent.Core/Interfaces/ProcessRunner.cs:77 in [b222b25](https://github.com/PlayFab/MpsAgent/commit/b222b2550f83e8ae8345fa6f5965f0d7755eb97b). [](commit_id = [b222b25](https://github.com/PlayFab/MpsAgent/commit/b222b2550f83e8ae8345fa6f5965f0d7755eb97b), deletion_comment = False)

It originally returned Task.CompletedTask and I was told to change it to throw, but I can change it back if it'll cause new issues

dgkanatsios:
> It originally returned Task.CompletedTask and I was told to change it to throw, but I can change it back if it'll cause new issues
Yeah, I can see the original comment. Can we make sure that it's not called internally by VmAgent?

yegarci:
> > It originally returned Task.CompletedTask and I was told to change it to throw, but I can change it back if it'll cause new issues
> > Yeah, I can see the original comment. Can we make sure that it's not called internally by VmAgent?

It's all good! No calls

VmAgent.Core/Interfaces/ProcessRunner.cs
```diff
@@ -79,6 +82,27 @@ public override Task CollectLogs(string processId, string logsFolder, ISessionHo
...
+        public override Task CreateStartGameExceptionLogs(string logsFolder, string exceptionMessage)
+        {
+            try
+            {
+                _logger.LogVerbose("Collecting logs for failed start game process.");
+                string destinationFileName = Path.Combine(logsFolder, ConsoleLogCaptureFileName);
+
+                using (StreamWriter sw = File.AppendText(destinationFileName))
+                {
+                    sw.WriteLine(DateTime.Now.ToString("HH:mm:ss D"));
```
emmayspark, line 94: I see the D as a character in the log file as below -
```
20:27:14 0:D
System.ComponentModel.Win32Exception (2): The system cannot find the file specified.
   at System.Diagnostics.Process.StartWithCreateProcess(ProcessStartInfo startInfo)
```
I think we can use (DateTime.Now.ToString("yyyy-MM-dd'T'HH:mm:ss.fffK") to match the date format in docker. 

Here is the example of logs from docker when it throws exception:
```
{"log":"Running Startup Script\n","stream":"stdout","time":"2022-06-15T21:37:43.0271122Z"}
{"log":"Certificates located at: C:\\GameCertificates\n","stream":"stdout","time":"2022-06-15T21:37:43.0481836Z"}
{"log":"'C:\\Assets\\wrappser.exe' is not recognized as an internal or external command,\r\n","stream":"stderr","time":"2022-06-15T21:37:43.25104Z"}
{"log":"operable program or batch file.\r\n","stream":"stderr","time":"2022-06-15T21:37:43.25104Z"}

```

emmayspark (changes requested):
Awesome Job on the first PR !!

emmayspark (approved):

PR #121, opened 2022-06-23
extending heartbeat response

dgkanatsios:
Extending heartbeat response object to include:
- full details of potential planned maintenance
- source of the maintenance event
- setting for titles that we enable this feature for

### DIFF
--- a/AgentInterfaces/MaintenanceSchedule.cs
+++ b/AgentInterfaces/MaintenanceSchedule.cs
@@ -19,10 +19,6 @@ namespace Microsoft.Azure.Gaming.AgentInterfaces
 
         public string ReportingVmId { get; set; }
 
-        /// <summary>
-        /// The source of the maintenance event. It can be either Azure or PlayFab
-        /// </summary>
-        public string MaintenanceSource { get; set; }
     }
 
     // https://docs.microsoft.com/en-us/azure/virtual-machines/windows/scheduled-events#query-for-events
--- a/AgentInterfaces/MaintenanceSchedule.cs
+++ b/AgentInterfaces/MaintenanceSchedule.cs
@@ -8,12 +8,6 @@ namespace Microsoft.Azure.Gaming.AgentInterfaces
     using Newtonsoft.Json;
     using Newtonsoft.Json.Converters;
 
-    [Serializable]
-    public class MaintenanceScheduleEx : MaintenanceSchedule
-    {
-        public MaintenanceReason MaintenanceReason { get; set; }
-    }
-
     // Data Format: https://docs.microsoft.com/en-us/azure/virtual-machines/windows/scheduled-events
     [Serializable]
     public class MaintenanceSchedule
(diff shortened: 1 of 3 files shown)


AgentInterfaces/MaintenanceSchedule.cs
```diff
@@ -38,4 +47,11 @@ public class MaintenanceEvent
 
         public int DurationInSeconds { get; set; }
     }
+
+    [JsonConverter(typeof(StringEnumConverter))]
+    public enum MaintenanceReason
```
ravarnamsft, line 52: I would call it source, I am not sure if we need an enum for it, we can just leave it as string

ravarnamsft: We can also add it to the MaintenanceSchedule itself, we may not need the additional class. I would call it MaintenanceSource

AgentInterfaces/MaintenanceSchedule.cs
```diff
@@ -16,9 +18,15 @@ public class MaintenanceSchedule
         public IList<MaintenanceEvent> MaintenanceEvents { get; set; }
 
         public string ReportingVmId { get; set; }
+
+        /// <summary>
+        /// The source of the maintenance event. It can be either Azure or PlayFab
+        /// </summary>
+        public string MaintenanceSource { get; set; }
```
brianwp3000, line 25: This field would mark the entire schedule as coming from either Azure or PlayFab--we should just use the existing EventSource field inside the MaintenanceEvent class to mark when an event comes from PlayFab

AgentInterfaces/VmAgentSettings.cs
```diff
@@ -21,6 +21,12 @@ public class VmAgentSettings
         /// </summary>
         public HashSet<string> TitlesUsingExternalAllocations { get; set; }
 
+        /// <summary>
+        /// Titles that are using MaintenanceV2 - essentially, getting the full maintenance schedule details on the
+        /// GSDK maintenance callback
+        /// </summary>
+        public HashSet<string> TitlesUsingMaintenanceV2 { get; set; }
```
brianwp3000, line 28: Instead of a thunderfig setting, shouldn't this be a flag that the GSDK sends to VmAgent? Like, as part of its first heartbeat can't it send a "usingMaintenanceV2" flag?

AgentInterfaces/SessionHostHeartbeatInfo.cs
```diff
@@ -59,6 +59,11 @@ public class SessionHostHeartbeatInfo
         /// </summary>
         public DateTime? NextScheduledMaintenanceUtc { get; set; }
 
+        /// <summary>
+        /// Schedule of the next planned maintenance
+        /// </summary>
+        public MaintenanceSchedule NextScheduledMaintenance { get; set; }
```
brianwp3000, line 65: The name "NextScheduledMaintenance" made more sense when it referred to the next upcoming maintenance event--I think the entire schedule should be called "ScheduledMaintenance" or "MaintenanceSchedule"

brianwp3000 (changes requested):

AgentInterfaces/MaintenanceSchedule.cs
dgkanatsios: ah, wasn't aware of that field, makes sense, will remove.

AgentInterfaces/SessionHostHeartbeatInfo.cs
dgkanatsios: assuming that we'll do the work of informing the user about all maintenance events (e.g. one generated from PlayFab and one from Azure - if they co-exist), I'm fine with this. Will amend.

pedrofv (approved):
:shipit:

brianwp3000 (approved):

PR #123, opened 2022-07-08
Added functionality to deploy from LMA

emmanuel-jr:
We want users to be able to create build locally and deploy to PlayFab Multiplayer Servers. With this implementation, users will be able to run `.\LocalMultiplayerAgent.exe -deploy` or `.\LocalMultiplayerAgent.exe -lcow -deploy` to deploy their build to PlayFab Multiplayer Servers. 

Added functionality to allow users to deploy their game servers through Local Multiplayer Agent:
- Added nuget packages for:
  - PlayFab C# SDK (for build API calls), 
  - Azure Blobs (for uploading files to blob storage)
  
- Added template `.json` file and corresponding class file (`DeploymentSettings.cs`) for additional parameters needed to run game server deployment
- Added `DeploymentSettingsValidator.cs` to validate `DeploymentSettings` parameters
- Added unit tests for validation of `DeploymentSettings` parameters; modified parameters to check if errors are actually caught by the `DeploymentSettingsValidator.cs` I created

- Added code to run deployment script (mainly in `DeploymentScript.cs`)
  - Added `Globals` `bool` parameter (`CreateDeployment`) to track user's choice to run game server deployment
  - Updated `MultiplayerSettingsValidator` to not check parameters not needed for game server deployment
  - Verified user authenticity by getting their Title SecretKey and EntityToken. Since user authentication is needed for using PlayFab APIs, users can dynamically enter their Title Secret Key in the console, or set it as a User Environment Variable (steps are in [documentation](https://github.com/PlayFab/MpsAgent/pull/134))
  - Checked if the following already exist and upload if not: `GameCertificates`, `LinuxImages`(for Linux Containers), `GameAssets`
  - Created requests and responses to create builds for Windows Process, Linux/Windows Container

Note: 
- TODO: Additional functional tests for `DeploymentScript.cs`


LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
```diff
@@ -0,0 +1,256 @@
+﻿using Microsoft.Azure.Gaming.LocalMultiplayerAgent.Config;
```
KalonJ-MSFT, line 1: nit: in general, favor alphabetizing these if you can. I don't know if we support doing system first or doing it all, but make sure you keep them in some order

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
```diff
@@ -0,0 +1,256 @@
...
+
+namespace Microsoft.Azure.Gaming.LocalMultiplayerAgent.MPSDeploymentTool
+{
+    public class DeploymentScript
+    {
+        private readonly MultiplayerSettings settings;
+        private readonly DeploymentSettings settingsDeployment;
+        public DeploymentScript(MultiplayerSettings multiplayerSettings)
+        {
+            settings = multiplayerSettings;
```
KalonJ-MSFT, line 21: recommendation: check these for null and throw early if they're not provided

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
```diff
@@ -0,0 +1,256 @@
...
+                    foreach (var file in request.GameAssetReferences)
+                    {
+                        await CheckAssetFiles(file.FileName);
+                    }
+
+                    createBuild = await CreateBuildWithManagedContainer(request);
+                }
+                else
+                {
+                    //for CustomLinux Container
```
KalonJ-MSFT, line 63: if this needs to be filled in later mark it as a TODO, I think - or do a throw unsupportedexception or something like that for now. Can always add it later

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
```diff
@@ -0,0 +1,256 @@
...
+                }
+            }
+        }
+
+        public CreateBuildWithCustomContainerRequest GetCustomContainerRequest()
+        {
+            return new CreateBuildWithCustomContainerRequest
+            {
+                BuildName = settingsDeployment.BuildName,
+                VmSize = (AzureVmSize)Enum.Parse(typeof(AzureVmSize), settingsDeployment.VmSize),
```
KalonJ-MSFT, line 96: If you do VmSize = Enum.Parse<AzureVmSize>(settingsDeployment.VmSize) here you can avoid the cast entirely and make it a strongly typed call.

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
```diff
@@ -0,0 +1,256 @@
...
+                },
+                Ports = PortMapping(),
+                ContainerRunCommand = settings.ContainerStartParameters.StartGameCommand,
+                RegionConfigurations = settingsDeployment.RegionConfigurations?.Select(x => new BuildRegionParams()
+                {
+                    Region = x.Region,
+                    MaxServers = x.MaxServers,
+                    StandbyServers = x.StandbyServers,
+                    MultiplayerServerCountPerVm = settingsDeployment.MultiplayerServerCountPerVm,
+                    VmSize = (AzureVmSize)Enum.Parse(typeof(AzureVmSize), settingsDeployment.VmSize)
```
KalonJ-MSFT, line 111: same goes here

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
```diff
@@ -0,0 +1,256 @@
...
+        public async Task<PlayFabResult<CreateBuildWithProcessBasedServerResponse>> CreateBuildWithProcessBasedServer(CreateBuildWithProcessBasedServerRequest request)
+        {
+            Console.WriteLine($"Starting deployment {request.BuildName} for titleId, regions  {string.Join(", ", request.RegionConfigurations.Select(x => x.Region))}");
+
+            return await PlayFabMultiplayerAPI.CreateBuildWithProcessBasedServerAsync(request);
+        }
+
+        public async Task<PlayFabResult<CreateBuildWithManagedContainerResponse>> CreateBuildWithManagedContainer(CreateBuildWithManagedContainerRequest request)
+        {
+            Console.WriteLine($"Starting deployment {request.BuildName} for titleId, regions  {string.Join(", ", request.RegionConfigurations.Select(x => x.Region))}");
```
KalonJ-MSFT, line 184: DRY (don't repeat yourself) this out - make a quickie method that logs deployment info and takes a request as a param

KalonJ-MSFT: And then use it in all the places

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
```diff
@@ -0,0 +1,256 @@
...
+            return await PlayFabMultiplayerAPI.CreateBuildWithCustomContainerAsync(request);
+        }
+
+        public List<PlayFab.MultiplayerModels.Port> PortMapping()
+        {
+            List<PlayFab.MultiplayerModels.Port> ports = null;
+
+            foreach (var portList in settings.PortMappingsList)
+            {
+                ports = portList?.Select(x => new PlayFab.MultiplayerModels.Port()
```
KalonJ-MSFT, line 202: this call will overwrite the ports variable every time so that only the last portlist in the portmappinglists gets put in. Is that intentional?

If it's not, what you probably want to do is something like

var ports = new List<PlayFab.MultiplayerModels.Port>();

ports.AddRange( portList?.Select(...))

That way you'll always return an empty list (which is usually good) and you'll be adding all the ports that apply here.

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
```diff
@@ -0,0 +1,256 @@
...
+                    Protocol = (ProtocolType)Enum.Parse(typeof(ProtocolType), x.GamePort.Protocol)
+                }).ToList();
+            }
+
+            return ports;
+        }
+
+        public string GetAssetFileNameFromPath(string filePath)
+        {
+            return System.IO.Path.GetFileName(filePath);
```
KalonJ-MSFT, line 215: nit: use a using statement to get rid of the preface for System.IO

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
```diff
@@ -0,0 +1,256 @@
...
+            var filevalidator = FileExistsInBlob(filename);
+
+            if (filevalidator.Result.Result == null)
+            {
+                GetAssetUploadUrlRequest request1 = new() { FileName = filename };
+
+                //TODO: log progress of asset upload
+                var uriResult = await PlayFabMultiplayerAPI.GetAssetUploadUrlAsync(request1);
+
+                if (uriResult.Error != null)
```
KalonJ-MSFT, line 237: If this is an error, do you want to continue to create a block blob and upload below?

LocalMultiplayerAgent/DeploymentTool/DeploymentSettings.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        public string BuildName { get; set; }
+
+        public string VmSize { get; set; }
+
+        public int MultiplayerServerCountPerVm { get; set; }
+
+        public string OSPlatform { get; set; }
+
+        public List<string> AssetFileNames { get; set; }
```
KalonJ-MSFT, line 19: A neat trick I like with Lists in Json is to declare them like this:

public List<string> AssetFileNames {get; } => new List<string>()

The reason is that that way this will never be null. Json deserializers know how to insert into a list. 

now, if this is getting passed in as a new List() by someone else, disregard - this is fine. 

(applies everywhere you use lists in json)

KalonJ-MSFT (changes requested):

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
```diff
@@ -0,0 +1,256 @@
...
+
+                foreach (var file in request.GameAssetReferences)
+                {
+                    await CheckAssetFiles(file.FileName);
+                }
+
+                createBuild = await CreateBuildWithProcessBasedServer(request);
+            }
+
+            if (createBuild.Error != null)
```
emmayspark, line 79: We can make a method that shows the error message from the response - same at line 39.

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
emmayspark, line 237: I think we want to check 
1) If asset exists - we can just go ahead 
2) if asset doesn't exist - we upload it
3) if we get errors from GetAssetUploadURLAsync - we show the logs something like "cannot get an URL to upload asset {Asset_Name}" with error message and return program.

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
```diff
@@ -0,0 +1,256 @@
...
+                //TODO: log progress of asset upload
+                var uriResult = await PlayFabMultiplayerAPI.GetAssetUploadUrlAsync(request1);
+
+                if (uriResult.Error != null)
+                {
+                    Console.WriteLine(uriResult.Error.ErrorMessage);
+                }
+                var uri = new System.Uri(uriResult.Result.AssetUploadUrl);
+
+                var blockBlob = new CloudBlockBlob(uri);
```
emmayspark, line 243: We also need to check the case when we put wrong URI so get error from blockBlob.UploadFromFileAsync(filename);
This case, we don't want to proceed because asset would not be uploaded to the blob storage so build cannot be created successfully.

emmayspark (changes requested):
Great Job on your first PR 👍🏼

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
emmanuel-jr: Wow....thank you for pointing this out @KalonJ-MSFT

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
KalonJ-MSFT: Sorry, it filtered out the lt and gt tags. It should be:

`VmSize = Enum.Parse<AzureVmSize>(settingsDeployment.VmSize)`

Let's see if that works

yegarci:
@emmanuel-jr Address failed build error and be sure to resolve conversations if you've fixed them in new commits

KalonJ-MSFT (approved):

LocalMultiplayerAgent/DeploymentTool/DeploymentSettingsValidator.cs
```diff
@@ -0,0 +1,102 @@
...
+        private readonly DeploymentSettings _settings;
+
+        public DeploymentSettingsValidator(DeploymentSettings settings)
+        {
+            _settings = settings ?? throw new ArgumentException("Deployment settings cannot be null");
+        }
+
+        public bool IsValid()
+        {
+            if (string.IsNullOrWhiteSpace(_settings.BuildName))
```
emmayspark, line 19: good job on validation check! It would be great if do validation check for VmSize and Region we support. 
For example, We support around 37 Region and 17 VmFamily (with different core option). We can separate file VmSizeInfo.cs and RegionInfo.cs and list out supported Region or VmFamily in there as Enum.

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
```diff
@@ -0,0 +1,270 @@
...
+        }
+
+        public async Task CheckAssetFiles(string filename)
+        {
+           
+            var filevalidator = FileExistsInBlob(filename);
+
+            if (filevalidator.Result.Result == null)
+            {
+                GetAssetUploadUrlRequest request1 = new() { FileName = filename };
```
emmayspark, line 247: make sure to replace new() to new GetAssetUploadUrlRequest () to fix an issue with test failure

emmayspark (changes requested):

LocalMultiplayerAgent/DeploymentTool/DeploymentSettingsValidator.cs
emmanuel-jr: Thanks for the hint! I made updates :)

emmanuel-jr:
I added the class for validation checks on new parameters added (`DeploymentSettingsValidator.cs`)
Feel free to review

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
```diff
@@ -0,0 +1,270 @@
...
+        }
+
+        public async Task RunScriptAsync()
+        { 
+            PlayFabSettings.staticSettings.TitleId = settings.TitleId;
+
+            string secret = Environment.GetEnvironmentVariable("PF_SECRET");
+            if (string.IsNullOrEmpty(secret))
+            {
+                Console.WriteLine("Enter developer secret key");
```
brianwp3000, line 32: We should notify the developer about what is happening here, like "Deploying to PlayFab Multiplayer Servers.... Enter developer secret key to continue:" or something

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
```diff
@@ -0,0 +1,270 @@
...
+                PlayFabSettings.staticSettings.DeveloperSecretKey = Console.ReadLine();
+            }
+
+            var req = new PlayFab.AuthenticationModels.GetEntityTokenRequest();
+            var res = await PlayFabAuthenticationAPI.GetEntityTokenAsync(req);
+
+            PrintError(res.Error);
+
+            // TODO: do validation checks
```
brianwp3000, line 41: it looks like you've added the validation so the TODO can be removed

LocalMultiplayerAgent/DeploymentTool/DeploymentSettingsValidator.cs
```diff
@@ -0,0 +1,107 @@
...
+
+namespace Microsoft.Azure.Gaming.LocalMultiplayerAgent.DeploymentTool
+{
+    public class DeploymentSettingsValidator
+    {
+        private readonly DeploymentSettings _settings;
+
+        public DeploymentSettingsValidator(DeploymentSettings settings)
+        {
+            _settings = settings ?? throw new ArgumentException("Deployment settings cannot be null");
```
KalonJ-MSFT, line 13: nit: in general favor using ArgumentNullException for checking null

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
```diff
@@ -0,0 +1,270 @@
...
+                Ports = PortMapping(),
+                ContainerRunCommand = settings.ContainerStartParameters.StartGameCommand,
+                RegionConfigurations = deploymentSettings.RegionConfigurations?.Select(x => new BuildRegionParams()
+                {
+                    Region = x.Region,
+                    MaxServers = x.MaxServers,
+                    StandbyServers = x.StandbyServers,
+                    MultiplayerServerCountPerVm = deploymentSettings.MultiplayerServerCountPerVm,
+                    VmSize = Enum.Parse<AzureVmSize>(deploymentSettings.VmSize)
+
```
brianwp3000, line 114: This is missing AssetReferences--Linux Container builds can use Assets as well

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
```diff
@@ -0,0 +1,270 @@
...
+                MultiplayerServerCountPerVm = deploymentSettings.MultiplayerServerCountPerVm,
+            };
+        }
+
+        public CreateBuildWithManagedContainerRequest GetManagedContainerRequest()
+        {
+            return new CreateBuildWithManagedContainerRequest
+            {
+                VmSize = Enum.Parse<AzureVmSize>(deploymentSettings.VmSize),
+                GameCertificateReferences = null,
```
brianwp3000, line 125: Is there a plan to enable developers to use this tool to upload certificates as well? Or if a developer is using certificates will they have to use the portal?

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
```diff
@@ -0,0 +1,270 @@
...
+            if (error != null && error.ErrorMessage != null)
+            {
+                Console.WriteLine($"{error.ErrorMessage}");
+            }
+        }
+
+        public async Task CheckAssetFiles(string filename)
+        {
+           
+            var filevalidator = FileExistsInBlob(filename);
```
brianwp3000, line 243: If I'm understanding this correctly, you're sending a request to ControlPlane to see if an asset already exists before attempting to upload it--could you instead just attempt to upload it and ControlPlane will return an error if the asset already exists?

LocalMultiplayerAgent/DeploymentTool/deployment.json
```diff
@@ -0,0 +1,16 @@
...
+    "VmSize": "Standard_D2_v2",
+    "MultiplayerServerCountPerVm": 10,
+    "RegionConfigurations": [
+        {
+            "Region": "EastUs",
+            "MaxServers": 10,
+            "StandbyServers": 5
+        }
+    ],
+    "OSPlatform": "Windows",
```
brianwp3000, line 12: I think we can deduce what OSPlatform the customer wants without this field. If it's a managed container based build, we know it's Windows. If it's a custom container-based server, we know it's Linux. If it's a process-based server, LMA can assume it's Windows because LMA doesn't support Linux process-based servers yet.
Having this field here just enables developers to run their Windows process-based servers in LMA and then deploy them as Linux process-based servers (which will not work out for them)

LocalMultiplayerAgent/DeploymentTool/deployment.json
```diff
@@ -0,0 +1,16 @@
...
+    "AssetFileNames": [
```
brianwp3000, line 13: Why can't we use the AssetDetails from MultiplayerSettings.json?

brianwp3000 (changes requested):

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
brianwp3000: Also, does this allow developer to reuse existing assets? Like, is there a way for them to create a build that uses assets they've already uploaded to a previous build?

LocalMultiplayerAgent/DeploymentTool/deployment.json
emmanuel-jr: Yes I was supposed to remove this one. I will do that in my next commit

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
emmanuel-jr: I added a try catch here, thanks!

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
emmanuel-jr: Yes, we can attempt to upload it and ControlPlane will return an error if the asset already exists. I've update. I'm also curious what your thinking is behind this. I'd think less async API calls is always better, i.e. more efficient, cheaper

brianwp3000: Fewer API calls is technically cheaper, but this is an extremely low traffic API--I'd be surprised if we got more than 100 asset uploads per day. And duplicating the validation logic that guarantees a duplicate Asset can't be uploaded just makes it harder to maintain down the road--if for some reason we changed how we validated that (maybe in the future we allow duplicate asset names?) we'd have 2 places that need changing.

emmanuel-jr:
I added new changes to upload certificates and check for Linux container images, please review. We currently don't have an API for uploading Linux container images so we let users know to upload before running deployment

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
```diff
@@ -0,0 +1,398 @@
...
+            
+            if (uriResult.Error != null)
+            {
+                if (uriResult.Error.ErrorMessage.Contains("AssetAlreadyExists"))
+                {
+                    Console.WriteLine($"{fileName} is already uploaded. Going ahead with deployment...");
+                }
+                else
+                {
+                    Console.WriteLine(uriResult.Error.ErrorMessage);
```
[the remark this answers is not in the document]
emmanuel-jr: Yeah I was just thinking about this; I'm not sure how to detect if the asset already in storage matches the asset they are trying to upload aside our `GetAssetUploadUrlAsync` API
I can add the prompt to user

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
```diff
@@ -0,0 +1,416 @@
...
+                    SkipToken = certSkipToken,
+                    PageSize = pageSize
+                };
+                var certificateSummariesResponse = await PlayFabMultiplayerAPI.ListCertificateSummariesAsync(certificateSummariesRequest);
+
+                ErrorCheck(certificateSummariesResponse.Error);
+
+                certSkipToken = certificateSummariesResponse.Result.SkipToken ?? null;
+
+                foreach (var certificate in settings.GameCertificateDetails)
```
emmayspark, line 303: I think we need to check if user configured cert information in GameCertificateDetails so we don't need to call ListCertificateSummariesAsync unnecessarily?

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
```diff
@@ -0,0 +1,416 @@
...
+
+            if (!validator.IsValid())
+            {
+                Console.WriteLine("The specified settings are invalid. Please correct them and re-run the agent.");
+                Environment.Exit(1);
+            }
+
+            await CheckAndUploadCertificatesAsync();
+
+            if (Globals.GameServerEnvironment == GameServerEnvironment.Windows)
```
emmayspark, line 47: users can upload game asset when they create Linux Build

LocalMultiplayerAgent/MultiplayerSettings.json
```diff
@@ -1,55 +1,56 @@
 {
-  "RunContainer": false,
```
emmayspark, line 2: I don't think we want to push this file ? Do you want to revert this?

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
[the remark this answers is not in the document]
emmanuel-jr: Feel free to review new changes

LocalMultiplayerAgent.UnitTest/DeploymentTests.cs
```diff
@@ -0,0 +1,119 @@
...
+        [TestMethod]
+        [TestCategory("BVT")]
+        public void InvalidBuildNameIsInvalid()
+        {
+            dynamic config = GetValidConfig();
+
+            DeploymentSettings settings = JsonConvert.DeserializeObject<DeploymentSettings>(config.ToString());
+            settings.BuildName = null;
+            
+            new DeploymentSettingsValidator(settings).IsValid().Should().BeFalse();
```
emmayspark, line 116: great job on adding unit tests!

LocalMultiplayerAgent/Program.cs
```diff
@@ -49,6 +48,14 @@ public static async Task Main(string[] args)
                 Environment.Exit(1);
             }
 
+            Globals.CreateDeployment = args.Contains("-deploy");
```
emmayspark, line 51: shouldn't this code run after debugging? how do we know from that args parameters if users just want to run deployment script or run both of debugging and deployment script?

My understanding of this logic is when args include -deploy, we don't do debugging and just run deploymentscript?

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
```diff
@@ -0,0 +1,493 @@
...
+            else
+            {
+                //call BuildSummariesAPI to confirm complete deployment ??
+                Console.WriteLine("Build creation was successful!");
+            }
+        }
+
+        public async Task<PlayFabResult<PlayFab.AuthenticationModels.GetEntityTokenResponse>> PlayFabAuthentication()
+        {
+            PlayFabSettings.staticSettings.TitleId = settings.TitleId;
```
emmayspark, line 110: We can write units test for the deployment script.
For example, we could create a unit test that checks when secret key is null of invalid, it throws exception properly.

LocalMultiplayerAgent.UnitTest/DeploymentTests.cs
```diff
@@ -0,0 +1,119 @@
...
+            //mockMultiplayerSettings.Setup(x => x.).Returns(true);
+            /*_mockSystemOperations.Setup(x => x.FileExists(It.IsAny<string>())).Returns(true);
+            _mockSystemOperations.Setup(x => x.DirectoryExists(It.IsAny<string>())).Returns(true);*/
+        }
+
+        [TestMethod]
+        [TestCategory("BVT")]
+        public void ValidConfigReturnsValid()
+        {
+            dynamic config = GetValidConfig();
```
emmayspark, line 53: These two lines below are repetitive in other tests so we can create a common method that includes these code.

 dynamic config = GetValidConfig();
 DeploymentSettings settings = JsonConvert.DeserializeObject<DeploymentSettings>(config.ToString());

LocalMultiplayerAgent.UnitTest/DeploymentTests.cs
```diff
@@ -0,0 +1,119 @@
...
+using Newtonsoft.Json.Linq;
+/*using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;*/
+
+namespace Microsoft.Azure.Gaming.LocalMultiplayerAgent.UnitTests
+{
+    [TestClass]
+    internal class DeploymentTests
```
emmayspark, line 16: nit: This is unit test for DeploymentSettings so we can rename it to DeploymentSettingsTests so not confused with DeploymentScriptTests

emmayspark (changes requested):

LocalMultiplayerAgent/Program.cs
emmanuel-jr: Yes, so if users want to run debugging before deployment, they run `.\LocalMultiplayerAgent.exe`. After this is ran, the program ends.
So to run deployment, the user will have to `.\LocalMultiplayerAgent.exe -deploy`, which runs the program again
They are two independent processes
Does that make sense?

LocalMultiplayerAgent.UnitTest/ConfigurationTests.cs
```diff
@@ -4,8 +4,6 @@
 namespace Microsoft.Azure.Gaming.VmAgent.UnitTests
 {
     using System;
-    using System.Configuration;
-    using System.IO;
```
emmanuel-jr, line 8: They are never used

LocalMultiplayerAgent/Config/MultiplayerSettings.cs
```diff
@@ -9,11 +9,9 @@ namespace Microsoft.Azure.Gaming.LocalMultiplayerAgent.Config
     using System.IO;
     using System.Linq;
     using System.Reflection;
-    using System.Runtime.InteropServices;
```
emmanuel-jr, line 12: Never used

LocalMultiplayerAgent/Globals.cs
```diff
@@ -20,6 +20,8 @@ public static class Globals
         // Value updated from settings.json
         public static SessionConfig SessionConfig = null;
 
+        public static bool CreateDeployment = false;
```
dgkanatsios, line 23: "Deployment" is an internal MPS name. "Build" is the name we expose to customers via our public APIs. I believe we should keep it consistent.

LocalMultiplayerAgent/Program.cs
```diff
@@ -37,9 +36,14 @@ public static async Task Main(string[] args)
             // lcow stands for Linux Containers On Windows => https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/linux-containers
             Globals.GameServerEnvironment = args.Contains("-lcow") && RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
                 ? GameServerEnvironment.Linux : GameServerEnvironment.Windows; // LocalMultiplayerAgent is running only on Windows for the time being
+            Globals.CreateDeployment = args.Contains("-deploy");
```
dgkanatsios, line 39: as mentioned before, I'd strongly recommend using the "Build" keyword instead of "Deploy"

LocalMultiplayerAgent/Program.cs
```diff
@@ -37,9 +36,14 @@ public static async Task Main(string[] args)
...
+
             MultiplayerSettings settings = JsonConvert.DeserializeObject<MultiplayerSettings>(File.ReadAllText("MultiplayerSettings.json"));
 
-            settings.SetDefaultsIfNotSpecified();
+            if (!Globals.CreateDeployment)
```
dgkanatsios, line 43: why is the variable global? can we have it as a local variable? Global variables are hard to test and error prone

LocalMultiplayerAgent/Config/MultiplayerSettingsValidator.cs
```diff
@@ -27,20 +27,9 @@ public MultiplayerSettingsValidator(MultiplayerSettings settings, ISystemOperati
 
             _systemOperations = systemOperations ?? SystemOperations.Default;
         }
-    
+
         public bool IsValid()
         {
-            if (string.IsNullOrWhiteSpace(_settings.OutputFolder) || string.IsNullOrWhiteSpace(_settings.TitleId))
```
dgkanatsios, line 33: why are we removing these from here? I can see that they are used below but they are both critical so I'd rather have the code fail early.

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
```diff
@@ -0,0 +1,496 @@
...
+using Microsoft.Azure.Storage.Blob;
+using Microsoft.Azure.Gaming.AgentInterfaces;
+using Microsoft.Azure.Gaming.LocalMultiplayerAgent.Config;
+using Newtonsoft.Json;
+using PlayFab;
+using PlayFab.MultiplayerModels;
+
+namespace Microsoft.Azure.Gaming.LocalMultiplayerAgent.MPSDeploymentTool
+{
+    public class DeploymentScript
```
dgkanatsios, line 19: As mentioned elsewhere, I'd rather call it "CreateBuildScript" or something

LocalMultiplayerAgent/Program.cs
dgkanatsios: FWIW, our APIs are named "CreateBuildXXX"

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
```diff
@@ -0,0 +1,496 @@
...
+
+        public DeploymentScript(MultiplayerSettings multiplayerSettings)
+        {
+            settings = multiplayerSettings ?? throw new ArgumentNullException(nameof(multiplayerSettings));
+            deploymentSettings = JsonConvert.DeserializeObject<DeploymentSettings>(File.ReadAllText("DeploymentTool/DeploymentSettings.json"));
+        }
+
+        public async Task RunScriptAsync()
+        {
+            Console.WriteLine("Deploying to PlayFab Multiplayer Servers...\n");
```
dgkanatsios, line 32: full name of the service is "Azure PlayFab Multiplayer Servers"

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
```diff
@@ -0,0 +1,496 @@
...
+
+            var auth = await PlayFabAuthentication();
+
+            ErrorCheck(auth.Error);
+
+            DeploymentSettingsValidator validator = new DeploymentSettingsValidator(deploymentSettings);
```
dgkanatsios, line 38: validating the deployment settings is faster than authenticating, I'd do that first :)

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
```diff
@@ -0,0 +1,496 @@
...
+                        {
+                            Console.WriteLine($"{errorMessage}");
+                        }
+                    }
+                    Console.WriteLine($"{createBuild.Error.ErrorMessage}");
+                }
+            }
+            else
+            {
+                //call BuildSummariesAPI to confirm complete deployment ??
```
dgkanatsios, line 103: is this a TODO ?

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
```diff
@@ -0,0 +1,496 @@
...
+                Console.WriteLine("Build creation was successful!");
+            }
+        }
+
+        public async Task<PlayFabResult<PlayFab.AuthenticationModels.GetEntityTokenResponse>> PlayFabAuthentication()
+        {
+            PlayFabSettings.staticSettings.TitleId = settings.TitleId;
+            var authValidation = new PlayFabResult<PlayFab.AuthenticationModels.GetEntityTokenResponse>();
+            string secret = Environment.GetEnvironmentVariable("PF_SECRET", EnvironmentVariableTarget.Process);
```
dgkanatsios, line 112: this is general comment, but it's useful to be verbose when doing CLI apps. For example, in this case, I'd use a Console.WriteLine("trying to read PF_SECRET environment variable"). If it fails to read it, Console.WriteLine("Failed to read PF_SECRET, enter developer secret key").
The golden rule of thumb is to never surprise the user :)

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
```diff
@@ -0,0 +1,496 @@
...
+        {
+            GetContainerRegistryCredentialsRequest requ = new GetContainerRegistryCredentialsRequest();
+            var ContainerRegistryCredentialsRes = await PlayFabMultiplayerAPI.GetContainerRegistryCredentialsAsync(requ);
+            ErrorCheck(ContainerRegistryCredentialsRes.Error);
+
+            try
+            {
+                DockerClient client = new DockerClientConfiguration().CreateClient();
+                string registryWithImageName = $"{imageDetails.Registry}/{imageDetails.ImageName}";
+                Progress<JSONMessage> theMess = new Progress<JSONMessage>(
```
dgkanatsios, line 274: good job on outputting the progress and informing the user!

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
```diff
@@ -0,0 +1,496 @@
...
+                    };
+                    var ContainerImageTagsResponse = await PlayFabMultiplayerAPI.ListContainerImageTagsAsync(ContainerImageTagsRequest);
+
+                    ErrorCheck(ContainerImageTagsResponse.Error);
+
+                    IEnumerable<string> existingImageTags = ContainerImageTagsResponse.Result.Tags.Where(x => x == settings.ContainerStartParameters.ImageDetails.ImageTag);
+
+                    if (existingImageTags.Any())
+                    {
+                        Console.WriteLine("Your container image is already uploaded. Continuing deployment...");
```
dgkanatsios, line 335: FWIW, some folks like to overwrite the tags especially during development. This is a (very) bad practice IMHO, good thing that we are now allowing it.

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
```diff
@@ -0,0 +1,496 @@
...
+        public async Task<PlayFabResult<EmptyResponse>> UploadCertificateAsync(GameCertificateDetails certificate)
+        {
+            X509Certificate2 certCopy = new X509Certificate2(certificate.Path);
+            var uploadCertificateRequest = new UploadCertificateRequest
+            {
+                GameCertificate = new Certificate
+                {
+                    Name = certificate.Name,
+                    Base64EncodedValue = Convert.ToBase64String(certCopy.RawData)
+                    //Password = ""  //Only passwordless certificates supported currently
```
dgkanatsios, line 357: is this a TODO? If it is, we should create an issue on the repository and link it here

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
```diff
@@ -0,0 +1,496 @@
...
+
+            return ports;
+        }
+
+        public async Task UploadAssetFileAsync(string uploadUrl, string fileNamePath)
+        {
+            string fileName = Path.GetFileName(fileNamePath);
+            try
+            {
+                //TODO: log progress of asset upload
```
dgkanatsios, line 434: for TODOs on open source projects, it's a good practice to create an issue on the repository and link it here

LocalMultiplayerAgent.UnitTest/DeploymentSettingsTests.cs
```diff
@@ -0,0 +1,111 @@
...
+        {
+            DeploymentSettings settings = GetTestDeploymentSettings();
+
+            settings.VmSize = settings.VmSize.ToLower();
+            new DeploymentSettingsValidator(settings).IsValid().Should().BeFalse();
+        }
+
+        [TestMethod]
+        [TestCategory("BVT")]
+        public void InValidRegionConfigIsInValid()
```
dgkanatsios, line 79: typo: lowercase V

LocalMultiplayerAgent.UnitTest/DeploymentSettingsTests.cs
```diff
@@ -0,0 +1,111 @@
...
+using Microsoft.Azure.Gaming.LocalMultiplayerAgent.DeploymentTool;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using System;
+
+namespace Microsoft.Azure.Gaming.LocalMultiplayerAgent.UnitTests
+{
+    [TestClass]
+    public class DeploymentSettingsTests
```
dgkanatsios, line 11: this class tests only for "false" outcomes, we should make sure that "true" outcomes work as well.

LocalMultiplayerAgent/Config/MultiplayerSettingsValidator.cs
emmanuel-jr: Ah okay I get it. It was basically because I was checking the `createDeployment` condition below, but yes I can move that up

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
emmanuel-jr: Mm..not a TODO. I was just considering calling the BuildSummariesAPI after creating the build to see if our newly created build is found, but I don't think it's necessary since we check for errors from the CreateBuild API

brianwp3000 (approved):

LocalMultiplayerAgent/DeploymentTool/DeploymentScript.cs
emmanuel-jr: Not necessarily a TODO, just a note to people (devs) so they know that `Password` is a possible parameter but we currently do not support that. I guess we could make it a TODO if we have customers request?
Is this confusing?

dgkanatsios: people that download LocalMultiplayerAgent via the "Releases" tab won't see this comment, since they download the code already compiled. The only way they can see it is if they check the LMA source code (in most cases they don't) or if they see an issue on the repository that this is a "known issue"

emmayspark: @emmanuel-jr  make sure to add this to the readme

emmayspark (approved):

PR #126, opened 2022-07-18
added GetEnvironmentVariablesForVmScripts

brianwp3000:
Currently, game servers have a PF_SHARED_CONTENT_FOLDER environment variable that points them to a folder shared among all servers on a VM. For process based servers that points to the actual folder, and for container based servers it points to where that folder is mounted in the container.
When running a VM startup script, however, that environment variable isn't set. This PR creates a new environment variable, PF_SHARED_CONTENT_FOLDER_VM that points to where the shared content folder is on the VM. The Watson team needs this for their startup script.
I put that new environment variable inside `GetEnvironmentVariablesForVmScripts` because it doesn't belong in the "common" environment variables (which end up inside container-based servers). There will be a separate VmAgent PR to make the startup script use these variables when it runs a startup script.

### DIFF
--- a/VmAgent.Core/VmConfiguration.cs
+++ b/VmAgent.Core/VmConfiguration.cs
@@ -37,6 +37,11 @@ namespace Microsoft.Azure.Gaming.VmAgent.Core
 
         public const string FqdnEnvVariable = "PF_FQDN";
 
+        // Used by the games to share user generated content (and other files that are downloaded once, used multiple times).
+        // This points to the folder on the VM, and should only be used by VM-level scripts. Game servers
+        // should use `PF_SHARED_CONTENT_FOLDER` instead (which points to wherever the folder is mounted in the container)
+        public const string SharedContentFolderVmVariable = "PF_SHARED_CONTENT_FOLDER_VM";
+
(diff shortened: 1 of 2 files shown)


VmAgent.Core/VmConfiguration.cs
```diff
@@ -98,6 +103,29 @@ public VmConfiguration(int listeningPort, string vmId, VmDirectories vmDirectori
...
         }
 
+        /// <summary>
+        /// Gets the set of environment variables for scripts running at the VM level.
+        /// </summary>
+        /// <param name="sessionHostsStartInfo">The details for starting the game servers.</param>
+        /// <param name="vmConfiguration">The details for the VM.</param>
+        /// <returns>A dictionary of environment variables</returns>
+        /// <remarks>This method is expected to be called only after the VM is assigned (i.e sessionHostsStartInfo is not null).</remarks>
+        public static IDictionary<string, string> GetEnvironmentVariablesForVmScripts(SessionHostsStartInfo sessionHostsStartInfo, VmConfiguration vmConfiguration)
```
dgkanatsios, line 113: nit: I'd call it "...ForVmStartupScripts" for consistency

dgkanatsios (approved):

PR #135, opened 2022-08-13
Add support for capturing the output of Processes

emmayspark:
This code supports capturing stderr and stdout logs of Processes. 
When Process object is created, Start Information is configured with event handlers to subscribe to events and redirect streams of standard output/errors to the event handlers. Whenever events are raised to event handlers, it writes them to the **PF_ConsoleLog.txt** until process is terminated or error is captured. 

Sample Output For Windows Process:
[PF_ConsoleLogs.txt](https://github.com/PlayFab/MpsAgent/files/9377768/PF_ConsoleLogs.txt)
Sample Output For Linux Process:
[PF_ConsoleLogs(Linux).txt](https://github.com/PlayFab/MpsAgent/files/9377801/PF_ConsoleLogs.Linux.txt)

### DIFF
--- a/VmAgent.Core.UnitTests/ProcessOutputLoggerTest.cs
+++ b/VmAgent.Core.UnitTests/ProcessOutputLoggerTest.cs
@@ -3,9 +3,11 @@ using FluentAssertions;
 using Microsoft.ApplicationInsights;
 using Microsoft.ApplicationInsights.Extensibility;
 using Microsoft.Azure.Gaming.VmAgent.Core;
+using Microsoft.Azure.Gaming.VmAgent.Core.Dependencies.Interfaces;
 using Microsoft.Azure.Gaming.VmAgent.Core.UnitTests;
 using Microsoft.Extensions.Logging.Abstractions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
 using System;
 using VmAgent.Core.Dependencies.Interfaces.Exceptions;
 
@@ -14,17 +16,32 @@ namespace VmAgent.Core.UnitTests
     [TestClass]
     public class ProcessOutputLoggerTest
     {
+        private string _validPath = "C:\\PF_Consolelog.txt";
         private MultiLogger _multiLogger;
+        private Mock<IFileWriteWrapper> _fileWriteWrapper;
+        private ProcessOutputLogger _processLogger;
 
         [TestInitialize]
         public void BeforeEachTest()
         {
             _multiLogger = new MultiLogger(NullLogger.Instance, new TelemetryClient(TelemetryConfiguration.CreateDefault()));
+            _fileWriteWrapper = new Mock<IFileWriteWrapper>();
+
+            Mock<ProcessOutputLogger> mockProcessLogger = new Mock<ProcessOutputLogger>(
+               _validPath,
+               _multiLogger,
+               _fileWriteWrapper.Object)
+            {
+                CallBase = true
+            };
+
+            _processLogger = mockProcessLogger.Object;
         }
 
         [TestMethod, TestCategory("BVT")]
         [DataRow(null)]
         [DataRow("")]
+        [DataRow(" ")]
         public void InvalidLogFileNameFailed(string invalidFilePath)
         {
             ExceptionAssert.Throws<ProcessOuputLoggerCreationFailedException>(() => new ProcessOutputLogger(invalidFilePath, _multiLogger));
@@ -37,5 +54,24 @@ namespace VmAgent.Core.UnitTests
             ProcessOutputLogger processLogger = new ProcessOutputLogger(validFilePath, _multiLogger);
             processLogger.GetProcessLogFilePath().Should().Equals(validFilePath);
         }
+
+        [TestMethod, TestCategory("BVT")]
+        public void FileWriteFailDoNotThrowException()
+        {
+            _fileWriteWrapper.Setup(x => x.Write(It.IsAny<string>())).Throws(new Exception());
+
+            Action act = () => _processLogger.Log("test123");
+            act.Should().NotThrow();
+        }
+
+        [TestMethod, TestCategory("BVT")]
+        public void FileCloseFailDoNotThrowException()
+        {
+            _fileWriteWrapper.Setup(x => x.Close()).Throws(new Exception());
+            
+            Action act = () => _processLogger.Close();
+            act.Should().NotThrow();
+        }
+
     }
 }
--- a/VmAgent.Core.UnitTests/ProcessOutputLoggerTest.cs
+++ b/VmAgent.Core.UnitTests/ProcessOutputLoggerTest.cs
@@ -2,22 +2,11 @@
 using FluentAssertions;
 using Microsoft.ApplicationInsights;
 using Microsoft.ApplicationInsights.Extensibility;
-using Microsoft.Azure.Gaming.VmAgent;
 using Microsoft.Azure.Gaming.VmAgent.Core;
-using Microsoft.Azure.Gaming.VmAgent.Core.Dependencies;
-using Microsoft.Azure.Gaming.VmAgent.Core.Interfaces;
 using Microsoft.Azure.Gaming.VmAgent.Core.UnitTests;
 using Microsoft.Extensions.Logging.Abstractions;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
-using Moq;
 using System;
-using System.Collections.Generic;
-using System.Diagnostics;
-using System.IO;
-using System.Linq;
-using System.Runtime.InteropServices;
-using System.Text;
-using System.Threading.Tasks;
 using VmAgent.Core.Dependencies.Interfaces.Exceptions;
 
 namespace VmAgent.Core.UnitTests
@@ -35,6 +24,8 @@ namespace VmAgent.Core.UnitTests
 
         [TestMethod, TestCategory("BVT")]
         [DataRow(null)]
+        [DataRow("")]
+        [DataRow("\\abc")]
         public void InvalidLogFileNameFailed(string invalidFilePath)
         {
             ExceptionAssert.Throws<ProcessOuputLoggerCreationFailedException>(() => new ProcessOutputLogger(invalidFilePath, _multiLogger));
@@ -46,6 +37,28 @@ namespace VmAgent.Core.UnitTests
         {
             ProcessOutputLogger processLogger = new ProcessOutputLogger(validFilePath, _multiLogger);
             processLogger.GetProcessLogFilePath().Should().Equals(validFilePath);
+        }
+
+        [TestMethod, TestCategory("BVT")]
+        public void CloseFileAfterCloseDoNotThrowException()
+        {
+            string fileFilePath = "C:\\PF_Consolelog.txt";
+            ProcessOutputLogger processLogger = new ProcessOutputLogger(fileFilePath, _multiLogger);
+            processLogger.Close();
+            
+            Action act = () => processLogger.Close();
+            act.Should().NotThrow();
+        }
+
+        [TestMethod, TestCategory("BVT")]
+        public void LogToFileAfterCloseFileDoNotThrowException()
+        {
+            string fileFilePath = "C:\\PF_Consolelog.txt";
+            ProcessOutputLogger processLogger = new ProcessOutputLogger(fileFilePath, _multiLogger);
+            processLogger.Close();
+            
+            Action act = () => processLogger.Log("ABCD");
+            act.Should().NotThrow();
         }
     }
 }
--- a/VmAgent.Core.UnitTests/ProcessOutputLoggerTest.cs
+++ b/VmAgent.Core.UnitTests/ProcessOutputLoggerTest.cs
@@ -58,7 +58,7 @@ namespace VmAgent.Core.UnitTests
         [TestMethod, TestCategory("BVT")]
         public void FileWriteFailDoNotThrowException()
         {
-            _fileWriteWrapper.Setup(x => x.Write(It.IsAny<string>())).Throws(new Exception());
+            _fileWriteWrapper.Setup(x => x.Write(It.IsAny<string>(), It.IsAny<string>())).Throws(new InvalidOperationException());
 
             Action act = () => _processLogger.Log("test123");
             act.Should().NotThrow();
--- a/VmAgent.Core.UnitTests/ProcessOutputLoggerTest.cs
+++ b/VmAgent.Core.UnitTests/ProcessOutputLoggerTest.cs
@@ -1,0 +1,51 @@
+﻿
+using FluentAssertions;
+using Microsoft.ApplicationInsights;
+using Microsoft.ApplicationInsights.Extensibility;
+using Microsoft.Azure.Gaming.VmAgent;
+using Microsoft.Azure.Gaming.VmAgent.Core;
+using Microsoft.Azure.Gaming.VmAgent.Core.Dependencies;
+using Microsoft.Azure.Gaming.VmAgent.Core.Interfaces;
+using Microsoft.Azure.Gaming.VmAgent.Core.UnitTests;
+using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Runtime.InteropServices;
+using System.Text;
+using System.Threading.Tasks;
+using VmAgent.Core.Dependencies.Interfaces.Exceptions;
+
+namespace VmAgent.Core.UnitTests
+{
+    [TestClass]
+    public class ProcessOutputLoggerTest
+    {
+        private MultiLogger _multiLogger;
+
+        [TestInitialize]
+        public void BeforeEachTest()
+        {
+            _multiLogger = new MultiLogger(NullLogger.Instance, new TelemetryClient(TelemetryConfiguration.CreateDefault()));
+        }
+
+        [TestMethod, TestCategory("BVT")]
+        [DataRow(null)]
+        public void InvalidLogFileNameFailed(string invalidFilePath)
+        {
+            ExceptionAssert.Throws<ProcessOuputLoggerCreationFailedException>(() => new ProcessOutputLogger(invalidFilePath, _multiLogger));
+        }
+
+        [TestMethod, TestCategory("BVT")]
+        [DataRow("C:\\PF_Consolelog.txt")]
+        public void ValidLogFileNameReturnFilePath(string validFilePath)
+        {
+            ProcessOutputLogger processLogger = new ProcessOutputLogger(validFilePath, _multiLogger);
+            processLogger.GetProcessLogFilePath().Should().Equals(validFilePath);
+        }
+    }
+}
--- a/VmAgent.Core.UnitTests/ProcessOutputLoggerTest.cs
+++ b/VmAgent.Core.UnitTests/ProcessOutputLoggerTest.cs
@@ -25,7 +25,6 @@ namespace VmAgent.Core.UnitTests
         [TestMethod, TestCategory("BVT")]
         [DataRow(null)]
         [DataRow("")]
-        [DataRow("\\abc")]
         public void InvalidLogFileNameFailed(string invalidFilePath)
         {
             ExceptionAssert.Throws<ProcessOuputLoggerCreationFailedException>(() => new ProcessOutputLogger(invalidFilePath, _multiLogger));
@@ -37,28 +36,6 @@ namespace VmAgent.Core.UnitTests
         {
             ProcessOutputLogger processLogger = new ProcessOutputLogger(validFilePath, _multiLogger);
             processLogger.GetProcessLogFilePath().Should().Equals(validFilePath);
-        }
-
-        [TestMethod, TestCategory("BVT")]
-        public void CloseFileAfterCloseDoNotThrowException()
-        {
-            string fileFilePath = "C:\\PF_Consolelog.txt";
-            ProcessOutputLogger processLogger = new ProcessOutputLogger(fileFilePath, _multiLogger);
-            processLogger.Close();
-            
-            Action act = () => processLogger.Close();
-            act.Should().NotThrow();
-        }
-
-        [TestMethod, TestCategory("BVT")]
-        public void LogToFileAfterCloseFileDoNotThrowException()
-        {
-            string fileFilePath = "C:\\PF_Consolelog.txt";
-            ProcessOutputLogger processLogger = new ProcessOutputLogger(fileFilePath, _multiLogger);
-            processLogger.Close();
-            
-            Action act = () => processLogger.Log("ABCD");
-            act.Should().NotThrow();
         }
     }
 }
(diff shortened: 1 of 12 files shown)


VmAgent.Core/Interfaces/ISessionHostRunner.cs
```diff
@@ -26,6 +26,5 @@ public interface ISessionHostRunner
 
         Task CollectLogs(string id, string logsFolder, ISessionHostManager sessionHostManager);
 
-        Task CreateStartGameExceptionLogs(string logFolder, string exceptionMessage);
```
emmayspark, line 29: this is not used in DockerContainerEngine.cs so I removed this and updated the one in ProcessRunner.cs. Maybe this method initially had a purpose. Let me know if I missed something.

VmAgent.Core/ProcessOutputLogger.cs
```diff
@@ -0,0 +1,97 @@
...
+                throw new ProcessOuputLoggerCreationFailedException($"LogFilePath cannot be null.");
+            }
+
+            try
+            {
+                _logWriter = new StreamWriter(File.OpenWrite(logFilePath));
+            }
+            catch (Exception ex)
+            {
+                _logger.LogInformation($"Exception was thrown while creating a Process Log file under {_logFilePath}. {ex}");
```
kerdrek, line 36: you are logging _logFilePath before it is set with the right value, just remove the _ and you are good to go

kerdrek (changes requested):

VmAgent.Core/Interfaces/ProcessRunner.cs
```diff
@@ -58,8 +69,10 @@ public override Task<SessionHostInfo> CreateAndStart(int instanceNumber, GameRes
 
             try
             {
-                string processId = _processWrapper.Start(processStartInfo).ToString();
-                sessionHostManager.UpdateSessionHostTypeSpecificId(sessionHostUniqueId, processId);
+                int processId = processOutputLogger != null ?
```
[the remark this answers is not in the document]
emmayspark: between line 48 - 56, I added code 
```
ProcessOutputLogger processOutputLogger = null;
            try
            {
                processOutputLogger = new ProcessOutputLogger(processOutputFilePathOnVm, _logger);
            }
            catch(Exception exception)
            {
                _logger.LogException($"Failed to create ProcessOutputLogger with instance number {instanceNumber}", exception);
            }
```

assumed there might be a case that processOutputLogger is not initialized

VmAgent.Core/ProcessOutputLogger.cs
emmayspark: good catch! will fix that.

VmAgent.Core.UnitTests/ProcessOutputLoggerTest.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        [TestMethod, TestCategory("BVT")]
+        [DataRow(null)]
+        public void InvalidLogFileNameFailed(string invalidFilePath)
+        {
+            ExceptionAssert.Throws<ProcessOuputLoggerCreationFailedException>(() => new ProcessOutputLogger(invalidFilePath, _multiLogger));
+        }
+
+        [TestMethod, TestCategory("BVT")]
+        [DataRow("C:\\PF_Consolelog.txt")]
```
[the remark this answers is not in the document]
emmayspark: I think Test with Linux Path would fail because it would try writing it on my local machine. I think to do this, we would need to change a design

VmAgent.Core/Interfaces/ISessionHostRunner.cs
yegarci: It was done like that to avoid breaking something in the future: https://github.com/PlayFab/MpsAgent/pull/112/files#r910292900
But seeing how you redesigned the method, it should be fine

yegarci:
Do you have an example text output for a run?

emmayspark:
@yegarci  updated the PR description - check out the link!

yegarci:
Sorry, I mean do you have an example output where the new stderror and stdout streams are written?

VmAgent.Core/ProcessOutputLogger.cs
```diff
@@ -0,0 +1,97 @@
...
+
+    public class ProcessOutputLogger: IProcessOutputLogger
+    {
+        private string _logFilePath;
+        private StreamWriter _logWriter;
+        private MultiLogger _logger;
+
+        public ProcessOutputLogger(string logFilePath, MultiLogger logger)
+        {
+            if (logFilePath == null)
```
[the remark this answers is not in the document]
KalonJ-MSFT: also, likely favor using IsNullOrWhitespace unless a whitespace path is acceptable

VmAgent.Core/ProcessOutputLogger.cs
```diff
@@ -0,0 +1,104 @@
...
+            _logger.LogInformation($"Process Log file is created under {_logFilePath}");
+        }
+
+        public void Log(string message)
+        {
+            if (_logWriter == null)
+            {
+                _logger.LogInformation($"StreamWriter is null or not created yet.");
+                return;
+            }
```
KalonJ-MSFT, line 52: How is this condition able to be hit?

VmAgent.Core/ProcessOutputLogger.cs
```diff
@@ -0,0 +1,104 @@
...
+            _logWriter.WriteLine($"{currentDate}\t{message}");
+            _logWriter.Flush();
+        }
+
+        public string GetProcessLogFilePath()
+        {
+            return _logFilePath;
+        }
+
+        public void Close()
```
KalonJ-MSFT, line 65: is close something that you want people to be able to call explicitly? If they have a closed object, should they be able to interact with it in any way? 

To me, this seems like a very standard Disposable/Using pattern, where you create something for a given scope and then it should go away outside of that scope. 

If that's not the case, this probably should have some kind of state that tracks whether or not the object is closed and throws an exception if they attempt to use it when it's closed. As far as I can tell if this was closed you would get a log saying 'streamwriter is null or not created yet' in the Log call, but you wouldn't get any error condition.

emmayspark: Thanks Kalon, good questions - My initial though was file should be open whne VmAgent starts running a new Process and file should not be exposed to users. File would be closed when Process is terminated because we need to capture all the output and errors of Process while it is running. Only Logger interacts with Process through event handlers. Do you think I missed something here?

KalonJ-MSFT: So there are a couple possibilities.

One is that Process implements IDisposable and has a disposable ProcessOutputLogger as part of its objects, and when it is disposed of it disposes of this object. That's probably the standard way of chaining disposable objects together. However, since ProcessRunner isn't disposable that's probably not what you want. 

So the next thing to do is make sure that this class enforces how it is supposed to act - which means that anyone who uses this class gets an error if they attempt to log if it is closed. What that means is either you have a separate variable to track if it is closed, or you use logWriter as a proxy for that, and whenever logwriter is null and you attempt to log you actually throw an exception instead of just logging. 

Another possible issue is that calling Close() is not always the right choice due to how the filesystem works - and if multiple things are logging/writing at once it might cause an issue. Generally you should call Dispose() instead of Close when dealing with streams, because they will do things like suppress finalizers and deal with other race conditions.

emmayspark: @KalonJ-MSFT  Thank you for the feedback - I will think about the way to track the opened files using LogWriter. I think we can keep the list of LogWriter and check regularly. For possible issue you mentioned,  I thought close() eventually calls dispose() in FileSystem - I will double check with that too.

KalonJ-MSFT: I think you're right and I misremebered - Close() calls Dispose():

[https://stackoverflow.com/questions/7524903/should-i-call-close-or-dispose-for-stream-objects](https://stackoverflow.com/questions/7524903/should-i-call-close-or-dispose-for-stream-objects)

So close is the right thing and I was wrong. Sorry!

yegarci (approved):

VmAgent.Core/ProcessOutputLogger.cs
emmayspark: I made an update so that it can throw an exception instead of just logging - exceptions would be thrown in wrapper class so it will not break any code. I was thinking we could manage the each filesystem object and monitor them like session hosts but it will be trickier and I don't come up with the cases when file is still open or alive after the process (game session) is terminated because file instance would be alive in the same scope as each Process where event handler is triggered from.

KalonJ-MSFT: I think this is a good place to end it. if it becomes a problem later we can revisit it then.

KalonJ-MSFT (approved):
:shipit:

kerdrek (approved):

VmAgent.Core/ProcessOutputLogger.cs
```diff
@@ -0,0 +1,98 @@
...
+                _logger.LogException($"Exception was thrown while closing a file..", ex);
+            }
+        }
+
+        public void ErrorOutputHandler(object sendingProcess, DataReceivedEventArgs outLine)
+        {
+            Process p = sendingProcess as Process;
+            if (p != null)
+            {
+                _logger.LogInformation($"ErrorOutputHandler was triggered - ProcessId: {p.Id}");
```
brianwp3000, line 86: Why do we log every time they write to stderr? There's nothing stopping a developer from using stderr to write messages just like they would on stdout, so this might write a ton of logs

VmAgent.Core/ProcessOutputLogger.cs
```diff
@@ -0,0 +1,98 @@
...
+            }
+
+            Log(outLine.Data);
+            Close();
```
brianwp3000, line 90: don't close here--this can be called multiple times

VmAgent.Core/Interfaces/ProcessRunner.cs
```diff
@@ -58,8 +69,10 @@ public override Task<SessionHostInfo> CreateAndStart(int instanceNumber, GameRes
 
             try
             {
-                string processId = _processWrapper.Start(processStartInfo).ToString();
-                sessionHostManager.UpdateSessionHostTypeSpecificId(sessionHostUniqueId, processId);
+                int processId = processOutputLogger != null ? 
+                    _processWrapper.StartWithEventHandler(processStartInfo, processOutputLogger.StdOutputHandler, processOutputLogger.ErrorOutputHandler) : _processWrapper.Start(processStartInfo);
```
brianwp3000, line 73: Both stdout and stderr are being written to the same file here, which might be fine but i think it would be better if they wrote to separate files. Currently there is no way to tell what is written to each stream--they both just look like text in the output file (in a cmd window you can tell because stderr text is usually written in a different color).
Also i don't know what considerations need to be taken about multithreading--if the game server writes to stdout and stderr at the same time, will the text get jarbled in the output file? Or will something worse happen like a crash on our end? If we have them write to separate files we avoid having to worry about all this

brianwp3000 (changes requested):

VmAgent.Core/Dependencies/FileWriteWrapper.cs
```diff
@@ -0,0 +1,64 @@
...
+    using System.Diagnostics;
+    using System.IO;
+    using ApplicationInsights;
+    using ApplicationInsights.DataContracts;
+    using global::VmAgent.Core.Dependencies.Interfaces.Exceptions;
+    using Microsoft.Azure.Gaming.VmAgent.Core.Dependencies.Interfaces;
+    using Microsoft.Extensions.Logging;
+    using Newtonsoft.Json;
+
+    public class FileWriteWrapper : IFileWriteWrapper
```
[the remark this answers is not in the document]
brianwp3000: Oh i see you and Kalon discussed this elsewhere--i'll defer to that decision

VmAgent.Core/ProcessOutputLogger.cs
emmayspark: Makes sense, the reason I put Log here is I wanted to check when the file is closed and though checking errors from Process could be helpful for debugging just in case file is not uploaded but yeah it can be messy

VmAgent.Core/ProcessOutputLogger.cs
emmayspark: what could be the case exception is thrown but process is still alive but we don't want to close a file?

VmAgent.Core/Interfaces/ProcessRunner.cs
```diff
@@ -68,9 +81,16 @@ public override Task<SessionHostInfo> CreateAndStart(int instanceNumber, GameRes
...
-                CreateStartGameExceptionLogs(logFolderPathOnVm, exceptionMessage);
-            }
 
+                if (processOutputLogger != null)
+                {
+                    CreateStartGameExceptionLogs(processOutputLogger, exceptionMessage);
+                }
+                else
+                {
+                    _logger.LogVerbose($"processOutputlogger is not initialized. Failed to write failed start game error logs to {ConsoleLogCaptureFileName} for Process.");
```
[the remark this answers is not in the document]
emmayspark: guess I missed it. good catch

VmAgent.Core/ProcessOutputLogger.cs
emmayspark: I was thinking about design that we close file from SessionHostManger but it could be trickier - is there any suggestion to detect process is terminated from event handler?

brianwp3000: I think you might be misunderstanding what stderr is--
So every process can write to both stdout and stderr. You typically write to stdout using `Console.WriteLine`, which is shorthand for `Console.Out.WriteLine`. In the same way, you can write to stderr using `Console.Error.WriteLine`. Writing to stderr isn't always associated with an Exception being thrown--it's just an output stream a process can write to. It just so happens that by default, uncaught Exceptions get written to stderr by default.
So this `ErrorOutputHandler` method will get called every time the process writes to stderr--not just when the process crashes.

emmayspark: I think I missed that point - then this case I should add event handler to `Process.Exited` so they catch errors when Process is terminated. will take a look

VmAgent.Core/Interfaces/ProcessRunner.cs
emmayspark: I think we want to keep the same log file as Container-based Game Server.  I like the idea that we separate the log files for different purposes but we might need to do the same thing for Container? I'm going through documents if writing a log to a same file from two different event handler would cause a problem.

VmAgent.Core/ProcessOutputLogger.cs
brianwp3000: It might not be necessary to do that--if we're writing stderr to a file, theoretically any uncaught exception in the process will get written to stderr and then written to the file. So when the process exits, the developer will have access to the uncaught exception in the output file.
I think this might even work for the "you entered an invalid start command" thing Yessenia's fix was for--the "that file or process wasn't found" error will get written to stderr and show up in your log file.
Have you tried testing this locally?

VmAgent.Core/Interfaces/ProcessRunner.cs
brianwp3000: If we want them in the same file, we should do something like what Docker does for their console output logs. Here's an example:
```
{"log":"Hello, World! \u003c- written to console before GSDK initialized\n","stream":"stdout","time":"2022-08-17T17:54:26.185099166Z"}
{"log":"{\n","stream":"stdout","time":"2022-08-17T17:54:26.305422455Z"}
{"log":"  \"certificateFolder\": \"/data/GameCertificates\",\n","stream":"stdout","time":"2022-08-17T17:54:26.305445455Z"}
{"log":"  \"buildId\": \"bc452904-c2f1-4acd-b068-4ddddfd1a34c\",\n","stream":"stdout","time":"2022-08-17T17:54:26.305451955Z"}
{"log":"  \"logFolder\": \"/data/GameLogs/\",\n","stream":"stdout","time":"2022-08-17T17:54:26.305456255Z"}
{"log":"  \"vmId\": \"vmss:EastUs:59F84_bdfab727-1382-47a0-be31-a960577004d0:c7ae0ca6-38bc-4774-805e-5bec765e6cdb\",\n","stream":"stdout","time":"2022-08-17T17:54:26.305460155Z"}
```
Each of these lines specifies that it was written to the "stdout" steam (and presumably would specify if something was written to the "stderr" stream as well). And if we're adding additional data to each line, we might as well add the datetime too, because that's probably useful as well

VmAgent.Core/ProcessOutputLogger.cs
emmayspark: failed process error is captured before process starts so we don't get errors from event handlers. This is the error when you put a wrong start command. If I understand correctly that fix was for showing the logs when it failed to start because GSDK logs files are not even generated. Correct me if I'm wrong!

``` 
2022-08-23T16:01:10.666-07:00	System.ComponentModel.Win32Exception (2): The system cannot find the file specified.
   at System.Diagnostics.Process.StartWithCreateProcess(ProcessStartInfo startInfo)
   at System.Diagnostics.Process.Start()
   at Microsoft.Azure.Gaming.VmAgent.Core.Interfaces.ProcessWrapper.StartWithEventHandler(ProcessStartInfo startInfo, Action`2 StdOutputHandler, Action`2 ErrorOutputHandler, Action`2 ProcessExitedHanlder) in D:\PlayFabGit\MpsAgent\VmAgent.Core\Interfaces\ProcessWrapper.cs:line 48
   at Microsoft.Azure.Gaming.VmAgent.Core.Interfaces.ProcessRunner.CreateAndStart(Int32 instanceNumber, GameResourceDetails gameResourceDetails, ISessionHostManager sessionHostManager) in D:\PlayFabGit\MpsAgent\VmAgent.Core\Interfaces\ProcessRunner.cs:line 72
   ```

File: VmAgent.Core/Interfaces/ProcessRunner.cs (hunk 10 above)
Comment (emmayspark) [reply VmAgent.Core/Interfaces/ProcessRunner.cs:73]: Thanks Brian, yes this is definitely more useful - I'll do research on how two ( or three if we add exit handler) event handler would write a file each time and share what approach would be the best

**Review by yegarci — APPROVED** [review]

File: VmAgent.Core/Interfaces/ProcessWrapper.cs [hunk 13]
```diff
@@ -35,6 +34,24 @@ public int Start(ProcessStartInfo startInfo)
...
+        public int StartWithEventHandler(ProcessStartInfo startInfo, Action<object, DataReceivedEventArgs> StdOutputHandler, Action<object, DataReceivedEventArgs> ErrorOutputHandler, Action<object, EventArgs> ProcessExitedHanlder)
+        {
+            startInfo.RedirectStandardOutput = true;
+            startInfo.RedirectStandardError = true;
+
+            Process process = new Process();
+            process.StartInfo = startInfo;
+            process.OutputDataReceived += new DataReceivedEventHandler(StdOutputHandler);
+            process.ErrorDataReceived += new DataReceivedEventHandler(ErrorOutputHandler);
+            process.Exited += new EventHandler(ProcessExitedHanlder);
```
brianwp3000, line 46: ProcessExitedHanlder -> ProcessExitedHandler (this typo appears in several places :p )

VmAgent.Core/Dependencies/FileWriteWrapper.cs
```diff
@@ -0,0 +1,64 @@
...
+            return false;
+        }
+
+        public void Write(string message, string streamType = "stdout")
+        {
+            if (_textWriter != null)
+            {
+                string currentDate = DateTime.Now.ToString("yyyy-MM-dd'T'HH:mm:ss.fffK");
+
+                _textWriter.WriteLine($"{{\"log\":\"{message}\", \"stream\":{streamType}, \"time\":{currentDate}}}");
```
brianwp3000, line 50: I think this will output a log that looks like this:
`{{"log":"Some message", "stream":stdout, "time":2022-09-29T00:00:00.000}}`
1) Why is it surrounded with `{{` instead of just `{`?
2) The `streamType` and `time` values both need to be surrounded by `"` to be proper json

VmAgent.Core/Dependencies/FileWriteWrapper.cs
```diff
@@ -0,0 +1,64 @@
...
+                return true;
+            }
+            return false;
+        }
+
+        public void Write(string message, string streamType = "stdout")
+        {
+            if (_textWriter != null)
+            {
+                string currentDate = DateTime.Now.ToString("yyyy-MM-dd'T'HH:mm:ss.fffK");
```
brianwp3000, line 48: use `DateTime.UtcNow.ToString("O")` instead of building the format string yourself

VmAgent.Core/Dependencies/FileWriteWrapper.cs
```diff
@@ -0,0 +1,64 @@
...
+        private TextWriter _textWriter;
+
+        public FileWriteWrapper()
+        {
+        }
+
+        public void CreateFile(string logFilePath)
+        {
+            _logFilePath = logFilePath;
+            _textWriter = TextWriter.Synchronized(new StreamWriter(File.OpenWrite(logFilePath)));
```
brianwp3000, line 29: That is fancy, i didn't know about `TextWriter.Synchronized`

VmAgent.Core/Dependencies/Interfaces/IFileWriteWrapper.cs
```diff
@@ -8,7 +8,7 @@ namespace Microsoft.Azure.Gaming.VmAgent.Core.Dependencies.Interfaces
     public interface IFileWriteWrapper 
     {
         public void CreateFile(string logFilePath);
-        public void Write(string message);
+        public void Write(string message, string streamType = "stdout");
```
brianwp3000, line 11: In general, default arguments in Interfaces aren't ideal. If you are going to have them, then you should at least get rid of the default arguments on the implementation.
The problem is that it's easy to create a bug when the interface and the implementation have different default values--it's not always intuitive which one the compiler is going to use.

brianwp3000 (changes requested):

VmAgent.Core/ProcessOutputLogger.cs
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        public void StdOutputHandler(object sendingProcess, DataReceivedEventArgs outLine)
+        {
+            Log(outLine.Data);
+        }
+
+        public void ProcessExitedHanlder(object sendingProcess, EventArgs outLine)
+        {
+            Log("Process is terminated.");
```
brianwp3000, line 93: I'm not sure this log is necessary. I guess it can be used to show that the process exited cleanly? The wording here can also lead to a bit of confusion--is the process terminating on its own or has it been terminated by an external force?

VmAgent.Core/Dependencies/FileWriteWrapper.cs
emmayspark: 1. {{ will be shown as { in txt and I did this to match the log file format with Docker. In PF_console.txt for Docker, each line has { } in it.
Ex: {"log":"Running Startup Script\n","stream":"stdout","time":"2022-06-15T21:35:47.5767715Z"}
2. gotcha

brianwp3000: Also I think you should call `JsonConvert.ToString(message)` to escape `message`--if  `message` contains characters like `"` or `{` it'll mess up the log line (I'm assuming we want the log line to be valid json--i don't think that's actually a requirement but it would probably be nice for developers)

brianwp3000 (approved):

PR #141, opened 2022-10-04
made PF_SHARED_CONTENT_FOLDER available to vm startup scripts

brianwp3000:
This fix is specifically for the Watson team. Previously we gave their startup script the PF_SHARED_CONTENT_FOLDER_VM variable so that their agent process knows where to find crash dumps, but it turns out their script also needs the PF_SHARED_CONTENT_FOLDER variable so that they can set the linux kernel core path correctly for container-based servers to place their dumps in the shared content folder.

So for example:
PF_SHARED_CONTENT_FOLDER_VM = /mnt/GameSharedContent <- this folder exists on the VM
PF_SHARED_CONTENT_FOLDER (on a process based server) = /mnt/GameSharedContent
PF_SHARED_CONTENT_FOLDER (on a container based server) = /data/GameSharedContent <- this folder exists only exists inside a container, and is where PF_SHARED_CONTENT_FOLDER_VM is mounted
Their startup script tells the Watson agent to look for crash dumps in PF_SHARED_CONTENT_FOLDER_VM. The linux kernel core path tells linux where to place crash dumps, so they set it to PF_SHARED_CONTENT_FOLDER_VM. Turns out this doesn't work for containers--when a process in a container crashed, it tried placing the crash dump into /mnt/GameSharedContent which is a path that doesn't exist inside the container.
But they can get it to work if they set the linux kernel core path to PF_SHARED_CONTENT_FOLDER--that way when a process inside a container crashed, it got placed in /data/GameSharedContent (which is mounted to /mnt/GameSharedContent like we want). And for process based servers PF_SHARED_CONTENT_FOLDER is set to /mnt/GameSharedContent like we want. 

tl;dr PF_SHARED_CONTENT_FOLDER_VM is the true path to the shared content folder on the VM, while PF_SHARED_CONTENT_FOLDER is the path where a server *thinks* the shared content folder is. Turns out both are useful for VM startup scripts to have

### DIFF
--- a/VmAgent.Core/VmConfiguration.cs
+++ b/VmAgent.Core/VmConfiguration.cs
@@ -39,14 +39,17 @@ namespace Microsoft.Azure.Gaming.VmAgent.Core
 
         // Used by the games to share user generated content (and other files that are downloaded once, used multiple times).
         // This points to the folder on the VM, and should only be used by VM-level scripts. Game servers
-        // should use `PF_SHARED_CONTENT_FOLDER` instead (which points to wherever the folder is mounted in the container)
+        // should use `PF_SHARED_CONTENT_FOLDER` instead.
         public const string SharedContentFolderVmVariable = "PF_SHARED_CONTENT_FOLDER_VM";
 
         // Used by the games to share user generated content (and other files that are downloaded once, used multiple times).
-        // Also available to VM-level scripts for special cases (e.g. Watson needs to set the Kernel core dump path to this path
-        // so that container-based servers place their dump files in the path available to the container)
-        private const string SharedContentFolderEnvVariable = "PF_SHARED_CONTENT_FOLDER";
-
+        // This is the variable that points to the shared content folder as seen by the game server.
+        // Because of this, it is the same as PF_SHARED_CONTENT_FOLDER_VM if the server is process-based and it is different if
+        // the server is container-based.
+        // Typically this wouldn't be needed by VM startup scripts (since it might point to a path that only exists inside a container)
+        // but the Watson team found an edge case where the VM needs access to it so we made it available to the VM startup scripts as well.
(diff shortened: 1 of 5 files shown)


VmAgent.Core/VmConfiguration.cs
```diff
@@ -42,6 +42,11 @@ public class VmConfiguration
         // should use `PF_SHARED_CONTENT_FOLDER` instead (which points to wherever the folder is mounted in the container)
         public const string SharedContentFolderVmVariable = "PF_SHARED_CONTENT_FOLDER_VM";
 
+        // Used by the games to share user generated content (and other files that are downloaded once, used multiple times).
```
dgkanatsios, line 45: can you add here the great explanation you put on the PR description? This will help the reader understand why we need two env variables.

VmAgent.Core/VmConfiguration.cs
```diff
@@ -42,6 +42,11 @@ public class VmConfiguration
...
+        // Also available to VM-level scripts for special cases (e.g. Watson needs to set the Kernel core dump path to this path
+        // so that container-based servers place their dump files in the path available to the container)
+        private const string SharedContentFolderEnvVariable = "PF_SHARED_CONTENT_FOLDER";
```
dgkanatsios, line 48: should we also add some tests to make sure the correct set of env variables are being set?

jotavalenciano (approved):

VmAgent.Core/VmConfiguration.cs
brianwp3000: I expanded the comment (but it's still not as long as the novel i wrote for the PR description)

VmAgent.Core/VmConfiguration.cs
brianwp3000: Added unit tests

dgkanatsios (approved):

PR #145, opened 2022-12-06
add VM Startup Script references for assignment

dgkanatsios:
This PR adds the necessary bits for the VM Startup Script reference, so it can be triggered upon VM Assignment.


pedrofv (approved):
:shipit:

emmayspark (approved):

PR #146, opened 2022-12-08
upgrading projects to .NET6

dgkanatsios:
This PR upgrades all projects to .NET6


jotavalenciano (approved):
Just concerned with the warnings. But let's create another work item to remove them

PR #150, opened 2023-01-12
Creating and listing containers via labels

dgkanatsios:
This PR modifies the underlying Docker APIs to 

- add a specific label to the created GameServer containers
- list the containers via these labels

### DIFF
--- a/VmAgent.Core/Interfaces/DockerContainerEngine.cs
+++ b/VmAgent.Core/Interfaces/DockerContainerEngine.cs
@@ -25,6 +25,10 @@ namespace Microsoft.Azure.Gaming.VmAgent.ContainerEngines
 
     public class DockerContainerEngine : BaseSessionHostRunner
     {
+        private const string ContainerLabelKey = "CREATED_BY";
+
+        private const string ContainerLabelValue = "PF_MPS_VMAGENT";
+        
         private const string DockerWindowsNamedPipe = "npipe://./pipe/docker_engine";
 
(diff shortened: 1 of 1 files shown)


VmAgent.Core/Interfaces/DockerContainerEngine.cs
```diff
@@ -613,7 +618,21 @@ public override async Task RetrieveResources(SessionHostsStartInfo sessionHostsS
 
         public override async Task<IEnumerable<string>> List()
         {
-            return (await _dockerClient.Containers.ListContainersAsync(new ContainersListParameters())).Select(x => x.ID);
+            return (await _dockerClient.Containers.ListContainersAsync(
```
ravarnamsft, line 621: We are sure this works fully well right? Because we are not putting this behind any flag.

ravarnamsft (approved):

VmAgent.Core/Interfaces/DockerContainerEngine.cs
ravarnamsft: I meant in the cases we are tackling today, where a network part or something gets created. I think this is clever, thanks for thinking of it. I would just make sure that this works for the case we have seen. If there are any logs that you want to add to VMAgent, go for it. 
For example, we can start listing all containers, logs ones that are created by us and others that aren't and then , in another PR, kill only the ones that we created.

dgkanatsios: chatted, we're good to go! thanks!

PR #159, opened 2023-03-28
unify GetImageNameAndTagFromContainerImageDetails

brianwp3000:
A while back i added logic to pull images using the image Digest instead of the image Tag. Turns out i messed up--we had duplicate code in two places and i only changed one of them.
There are two places we need to specify an Tag/Digest:
1) when we pull the image from the remote registry
2) when we create a container using an image that we've already pulled

I only updated the 2nd place. So right now we are continuing to pull images from the remote registry using their tag, and then we use their digest to create containers with them. This works because the image referenced by the tag should also have the digest that we expect. But this opens us up to issues where the image referenced by a tag DOESN'T have the digest that we expect--in that case, we pull image with tag A and then try to create a container with Digest B and can't find the image that we pulled.

In practice this should never happen, but i'm unifying the logic so that now everywhere uses the same logic for identifying what image to use: `GetImageNameAndTagFromContainerImageDetails()`.

Also, using the digest isn't necessary anymore--i did that as part of the consolidatedAcr work and it ended up not being necessary.

### DIFF
--- a/VmAgent.Core.UnitTests/DockerContainerEngineTests.cs
+++ b/VmAgent.Core.UnitTests/DockerContainerEngineTests.cs
@@ -24,27 +24,10 @@ namespace VmAgent.Core.UnitTests
                 ImageDigest = null,
                 Registry = null,
             };
-            Assert.AreEqual("name:tag", DockerContainerEngine.GetImageNameFromContainerImageDetails(imageDetails));
+            Assert.AreEqual("name:tag", DockerContainerEngine.GetImageNameAndTagFromContainerImageDetails(imageDetails));
 
             imageDetails.Registry = "registry";
-            Assert.AreEqual("registry/name:tag", DockerContainerEngine.GetImageNameFromContainerImageDetails(imageDetails));
-        }
-
-        [TestMethod]
-        [TestCategory("BVT")]
-        public void TestGetImageNameWithDigest()
-        {
-            var imageDetails = new ContainerImageDetails
-            {
-                ImageName = "name",
-                ImageTag = "tag",
-                ImageDigest = "digest",
-                Registry = null,
-            };
-            Assert.AreEqual("name@digest", DockerContainerEngine.GetImageNameFromContainerImageDetails(imageDetails));
-
-            imageDetails.Registry = "registry";
-            Assert.AreEqual("registry/name@digest", DockerContainerEngine.GetImageNameFromContainerImageDetails(imageDetails));
+            Assert.AreEqual("registry/name:tag", DockerContainerEngine.GetImageNameAndTagFromContainerImageDetails(imageDetails));
         }
     }
 }
--- a/VmAgent.Core/Interfaces/DockerContainerEngine.cs
+++ b/VmAgent.Core/Interfaces/DockerContainerEngine.cs
@@ -415,7 +415,7 @@ namespace Microsoft.Azure.Gaming.VmAgent.ContainerEngines
             return sessionHost;
         }
 
-        public static string GetImageNameFromContainerImageDetails(ContainerImageDetails imageDetails)
+        private static string GetImageNameFromContainerImageDetails(ContainerImageDetails imageDetails)
         {
             if (imageDetails == null)
             {
@@ -433,7 +433,7 @@ namespace Microsoft.Azure.Gaming.VmAgent.ContainerEngines
             return imageName;
         }
 
-        public static string GetTagFromContainerImageDetails(ContainerImageDetails imageDetails)
+        private static string GetTagFromContainerImageDetails(ContainerImageDetails imageDetails)
         {
             if (imageDetails == null)
             {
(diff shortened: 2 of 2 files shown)


VmAgent.Core/Interfaces/DockerContainerEngine.cs
```diff
@@ -441,6 +433,26 @@ public static string GetImageNameFromContainerImageDetails(ContainerImageDetails
             return imageName;
         }
 
+        public static string GetTagFromContainerImageDetails(ContainerImageDetails imageDetails)
```
jotavalenciano, line 436: I think this function and the one above can be private

VmAgent.Core/Interfaces/DockerContainerEngine.cs
```diff
@@ -415,22 +415,14 @@
...
-                imageName = $"{imageDetails.ImageName}@{imageDetails.ImageDigest}";
-            }
-            else
-            {
-                imageName = $"{imageDetails.ImageName}:{imageDetails.ImageTag ?? "latest"}";
-            }
+            string imageName = imageDetails.ImageName;
 
             // Support running local images with no explicit registry in the name.
             if (!string.IsNullOrEmpty(imageDetails.Registry))
```
dgkanatsios, line 428: IIRC, this was to support "local" images being built locally and consumed locally by LMA. Does this work now with this change?

VmAgent.Core/Interfaces/DockerContainerEngine.cs
```diff
@@ -556,31 +568,30 @@
```
dgkanatsios: this was here before, but in any case this is not an accurate comment. We are not deleting the image (docker rmi command) but rather the instance of the image (the actual container), so I'd remove the "image" from the log.

VmAgent.Core/Interfaces/DockerContainerEngine.cs
```diff
@@ -415,22 +415,14 @@
             return sessionHost;
         }
 
-        public static string GetImageNameFromContainerImageDetails(ContainerImageDetails imageDetails)
```
dgkanatsios, line 418: >GetImageNameFromContainerImageDetails [](http://example.com/codeflow?start=29&length=37)

Does the image name contain the registry as well?

dgkanatsios: yes it does, ignore

VmAgent.Core/Interfaces/DockerContainerEngine.cs
brianwp3000: Short answer: yes.

Long answer:
When pulling images from the remote registry, prior to this change we assumed that the registry was specified and always appended it to the image name (so if the registry is null we would attempt to pull "<null>/myImage:tag"). When creating containers locally, we had logic to only append the registry if it was specified put in place for LocalMultiplayerAgent (so if the registry is null we attempt to create a container using image "myImage:tag"). Now with my change, since the logic is unified in both places we will attempt to use image "myImage:tag" instead of "<null>/myImage:tag" if the registry is null.

dgkanatsios: Thanks. Let's do a LocalMultiplayerAgent release when we merge this, just to make sure!

dgkanatsios (approved):

VmAgent.Core/Interfaces/DockerContainerEngine.cs
brianwp3000: The output of this method will contain the registry name as well

VmAgent.Core.UnitTests/DockerContainerEngineTests.cs
```diff
@@ -24,27 +24,10 @@ public void TestGetImageNameWithTag()
...
-                ImageName = "name",
-                ImageTag = "tag",
-                ImageDigest = "digest",
-                Registry = null,
-            };
-            Assert.AreEqual("name@digest", DockerContainerEngine.GetImageNameFromContainerImageDetails(imageDetails));
-
-            imageDetails.Registry = "registry";
-            Assert.AreEqual("registry/name@digest", DockerContainerEngine.GetImageNameFromContainerImageDetails(imageDetails));
+            Assert.AreEqual("registry/name:tag", DockerContainerEngine.GetImageNameAndTagFromContainerImageDetails(imageDetails));
```
yegarci, line 30: Would it be worthwhile to also check the individual private methods results equal the new combo method?

yegarci (approved):

VmAgent.Core.UnitTests/DockerContainerEngineTests.cs
brianwp3000: Nah I'm not too worried about it--if someone changes one of those private methods, this test will fail and we'll figure it out

PR #162, opened 2023-04-17
Add Multiple IP Addresses to VM Configuration

jotavalenciano:
Add the property and environment variables related to Multiple IP Addresses per game server

### DIFF
--- a/AgentInterfaces/SessionHostsStartInfo.cs
+++ b/AgentInterfaces/SessionHostsStartInfo.cs
@@ -5,7 +5,10 @@ namespace Microsoft.Azure.Gaming.AgentInterfaces
 {
     using System;
     using System.Collections.Generic;
+    using System.Diagnostics.Contracts;
     using Docker.DotNet.Models;
+    using Newtonsoft.Json.Converters;
+    using Newtonsoft.Json;
 
     public class SessionHostsStartInfo
     {
@@ -58,6 +61,11 @@ namespace Microsoft.Azure.Gaming.AgentInterfaces
         /// The fully qualified domain name for the Vm.
         /// </summary>
         public string FQDN { get; set; }
+
+        /// <summary>
+        /// List of Public IP Addresses
+        /// </summary>
+        public List<PublicIpAddress> PublicIpAddresses { get; set; }
 
         /// <summary>
         /// Indicates how the game server is hosted on the VM.
@@ -246,5 +254,24 @@ namespace Microsoft.Azure.Gaming.AgentInterfaces
         public int NodePort { get; set; }
         // TCP or UDP
         public string Protocol { get; set; }
+    }
+
+    [JsonConverter(typeof(StringEnumConverter))]
+    public enum RoutingType
+    {
+        Microsoft,
+        Internet
+    }
+
+    public class PublicIpAddress
+    {
+        // Frontend IP Address
+        public string IpAddress { get; set; }
+
+        // FQDN for the Public IP
+        public string FQDN { get; set; }
+
+        // Public IP Routing Type (Microsoft or Internet Routing)
+        public RoutingType RoutingType { get; set;}
     }
 }
--- a/AgentInterfaces/VmNetworkConfiguration.cs
+++ b/AgentInterfaces/VmNetworkConfiguration.cs
@@ -24,6 +24,11 @@
     public class Endpoint
     {
         /// <summary>
+        /// The frontend IP address for the endpoint.
+        /// </summary>
+        public string FrontendIpAddress { get; set; }
+
+        /// <summary>
         /// The publicly accessible port exposed on the software load balancer.
         /// </summary>
         public int FrontEndPort { get; set; }
--- a/VmAgent.Core.UnitTests/VmConfigurationTests.cs
+++ b/VmAgent.Core.UnitTests/VmConfigurationTests.cs
@@ -98,6 +98,37 @@ namespace VmAgent.Core.UnitTests
             }
         }
 
+        [TestMethod]
+        [TestCategory("BVT")]
+        public void EnvironmentVariablesWithMultipleIpAddresses()
+        {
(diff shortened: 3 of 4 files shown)


VmAgent.Core.UnitTests/VmConfigurationTests.cs
```diff
@@ -98,19 +98,50 @@ public void VmStartupScriptEnvVariablesWithPorts()
...
+                new PublicIpAddress()
+                {
+                    IpAddress = "127.0.0.2",
+                    FQDN = "myFQDN2",
+                    RoutingType = RoutingType.Microsoft
+                },
+            };
+
+            IDictionary<string, string> envVariables = VmConfiguration.GetCommonEnvironmentVariables(sessionHostsStartInfo, VmConfiguration);
+            envVariables.Should().Contain("PF_PUBLIC_IP_ADDRESSES_COUNT", "2");
```
emmayspark, line 123: I think we can make a const string for PF_PUBLIC_IP_ADDRESSES and create a helper function where tests env variables?

AgentInterfaces/VmNetworkConfiguration.cs
```diff
@@ -23,6 +23,11 @@ public class VmNetworkConfiguration
 
     public class Endpoint
     {
+        /// <summary>
+        /// The frontend IP address for the endpoint.
+        /// </summary>
+        public string FrontendIpAddress { get; set; }
```
emmayspark, line 29: where is this used?

jotavalenciano: This will be used in the VMAgent to get the different endpoints. It was not used before, but I will from now on

VmAgent.Core.UnitTests/VmConfigurationTests.cs
jotavalenciano: I already did that in code. Personally, I don't like to mix the usage of test and src code. If there is a code change, I want to see that change in tests as well.

AgentInterfaces/SessionHostsStartInfo.cs
```diff
@@ -59,6 +62,11 @@ public class SessionHostsStartInfo
         /// </summary>
         public string FQDN { get; set; }
 
+        /// <summary>
+        /// List of Public IP Addresses
+        /// </summary>
+        public List<PublicIpAddress> PublicIpAddresses { get; set; }
```
dgkanatsios, line 68: this will contain the one in the PublicIpV4Address, right?

dgkanatsios: (just curious)

VmAgent.Core/VmConfiguration.cs
```diff
@@ -49,7 +50,10 @@ public class VmConfiguration
         // Typically this wouldn't be needed by VM startup scripts (since it might point to a path that only exists inside a container)
         // but the Watson team found an edge case where the VM needs access to it so we made it available to the VM startup scripts as well.
         public const string SharedContentFolderEnvVariable = "PF_SHARED_CONTENT_FOLDER";
-        
+
+        /// Prefix used for all the Public Ip Addresses environment variables (count, ip addresss, fqdn, routing type)
+        public const string PublicIpAddressesEnvVariablePrefix = "PF_PUBLIC_IP_ADDRESSES";
```
dgkanatsios, line 55: not strong feelings, maybe add the _ here so the variables later contain only the actual name? But it's OK not to

VmAgent.Core.UnitTests/VmConfigurationTests.cs
dgkanatsios: yeah, I agree with Jota, let's leave it like this since tests sometimes are used as documentation so it's great to have the entire string here.

AgentInterfaces/VmNetworkConfiguration.cs
dgkanatsios: Where is this gonna used? Just curious

dgkanatsios (approved):
:shipit:

AgentInterfaces/VmNetworkConfiguration.cs
emmayspark: I had a same question, Jota mentioned that it will be used in VM Agent, maybe we should add a comment?

emmayspark (approved):

AgentInterfaces/SessionHostsStartInfo.cs
jotavalenciano: Correct. I will contain all of them. If there is only one, it will use the PublicIpV4Address one

AgentInterfaces/SessionHostsStartInfo.cs
```diff
@@ -247,4 +255,23 @@ public class VmStartupScriptPort
...
+    public enum RoutingType
+    {
+        Microsoft,
+        Internet
+    }
+
+    public class PublicIpAddress
+    {
+        // Frontend IP Address
+        public string IpAddress { get; set; }
```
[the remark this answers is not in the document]
jotavalenciano: I thought the same. But I would use another property with IpType. If we have Ipv4Address I don't want to add another field with Ipv6Address and make the user look for one or the other one. So I prefer to leave IpAddress and then in the future we can add IpType if it is Ipv4 or Ipv6

VmAgent.Core/VmConfiguration.cs
jotavalenciano: I tried it, but TBH it doesn't look great once you use it in code with the rest. I would prefer to leave it as it is. But thanks for the suggestion.

AgentInterfaces/SessionHostsStartInfo.cs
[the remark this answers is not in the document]
dgkanatsios: > So I prefer to leave IpAddress and then in the future we can add IpType if it is Ipv4 or Ipv6

what if we enable both IPv4 and IPv6? We're gonna need two fiels so Rakesh's comment makes sense.

PR #168, opened 2023-06-14
Add Logger For OpenTelemetry

emmayspark:
Logger will be passed by VmAgent. This is the first step to enable OpenTelemetry.

### DIFF
--- a/VmAgent.Core/MultiLogger.cs
+++ b/VmAgent.Core/MultiLogger.cs
@@ -21,9 +21,9 @@ namespace Microsoft.Azure.Gaming.VmAgent.Core
         {
             _logger = logger ?? throw new ArgumentException("logger cannot be null");
 
-			// To enable both GenevaAgentChannel and OpenTelemetry, we have 2 ILogger parameters here.
+            // To enable both GenevaAgentChannel and OpenTelemetry, we have 2 ILogger parameters here.
             // We will replace GenevaAgentchannel with Opentelemetry, once we verify all logs both in test and prod environments.
-			_genevaTelemetryClient = genevaTelemetryClient ?? throw new ArgumentException("genevaTelemetryClient cannot be null");
+            _genevaTelemetryClient = genevaTelemetryClient ?? throw new ArgumentException("genevaTelemetryClient cannot be null");
             _genevaOTelLogger = genevaLogger;
         }
 
--- a/VmAgent.Core/MultiLogger.cs
+++ b/VmAgent.Core/MultiLogger.cs
@@ -13,23 +13,23 @@ namespace Microsoft.Azure.Gaming.VmAgent.Core
     public class MultiLogger : ILogger
     {
         private readonly ILogger _logger;
-		private readonly TelemetryClient _genevaTelemetryClient;
-		private ILogger _genevaOTelLogger;
+        private readonly TelemetryClient _genevaTelemetryClient;
+        private ILogger _genevaOTelLogger;
         private const string VmAgentLoggerName = "VmAgentLogger";
 
-		public MultiLogger(ILogger logger, TelemetryClient genevaTelemetryClient, ILogger genevaLogger = null)
+        public MultiLogger(ILogger logger, TelemetryClient genevaTelemetryClient, ILogger genevaLogger = null)
         {
             _logger = logger ?? throw new ArgumentException("logger cannot be null");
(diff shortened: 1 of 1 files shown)


VmAgent.Core/MultiLogger.cs
```diff
@@ -13,62 +13,74 @@ namespace Microsoft.Azure.Gaming.VmAgent.Core
     public class MultiLogger : ILogger
     {
         private readonly ILogger _logger;
-        private readonly TelemetryClient _genevaTelemetryClient;
+		private readonly TelemetryClient _genevaTelemetryClient;
+		private ILogger _genevaOTelLogger;
+        private const string VmAgentLoggerName = "VmAgentLogger";
 
-        public MultiLogger(ILogger logger, TelemetryClient genevaTelemetryClient)
+		public MultiLogger(ILogger logger, TelemetryClient genevaTelemetryClient, ILogger genevaLogger = null)
```
pedrofv, line 20: >MultiLogger [](http://example.com/codeflow?start=9&length=11)

We should add comments on why we have 2 ILogger parameters here.

Are we going to remove/replace the existing ILogger in the future?

pedrofv (approved):
:shipit:

VmAgent.Core/MultiLogger.cs
```diff
@@ -13,62 +13,74 @@ namespace Microsoft.Azure.Gaming.VmAgent.Core
...
-        }
+			_genevaTelemetryClient.TrackTrace(message, SeverityLevel.Error);
+			_genevaOTelLogger?.LogError(message);
+		}
 
         public void LogException(Exception exception)
         {
             string message = exception.ToString();
             _logger.LogError(message);
-            _genevaTelemetryClient.TrackTrace(message, SeverityLevel.Error);
```
yegarci, line 51: Why is this _genevaTelemetryClient deleted, but the rest stayed?

yegarci (changes requested):

yegarci (approved):

VmAgent.Core/MultiLogger.cs
```diff
@@ -14,11 +14,22 @@ public class MultiLogger : ILogger
     {
         private readonly ILogger _logger;
         private readonly TelemetryClient _genevaTelemetryClient;
+        private ILogger _genevaOTelLogger;
```
dgkanatsios, line 17: nit: _genevaOtelLogger (lowercase T?)

VmAgent.Core/MultiLogger.cs
```diff
@@ -14,11 +14,22 @@ public class MultiLogger : ILogger
...
         {
             _logger = logger ?? throw new ArgumentException("logger cannot be null");
+
+            // To enable both GenevaAgentChannel and OpenTelemetry, we have 2 ILogger parameters here.
+            // We will replace GenevaAgentchannel with Opentelemetry, once we verify all logs both in test and prod environments.
             _genevaTelemetryClient = genevaTelemetryClient ?? throw new ArgumentException("genevaTelemetryClient cannot be null");
+            _genevaOTelLogger = genevaLogger;
+        }
+
+        public void SetUpLogger(ILoggerFactory loggerFactory)
```
dgkanatsios, line 30: why do we have the separate method if we're setting the same variable on the constructor?

VmAgent.Core/MultiLogger.cs
```diff
@@ -14,11 +14,16 @@ public class MultiLogger : ILogger
...
 
-        public MultiLogger(ILogger logger, TelemetryClient genevaTelemetryClient)
+        public MultiLogger(ILogger logger, TelemetryClient genevaTelemetryClient, ILogger genevaOpenTelemetryLogger = null)
         {
             _logger = logger ?? throw new ArgumentException("logger cannot be null");
+
+            // We have 2 ILogger parameters (1. Logger to write logs to local file. 2.Geneva OpenTelemetry Logger)
+            // We will replace GenevaAgentchannel with OpenTelemetry, once we verify all logs both in test and prod environments.
             _genevaTelemetryClient = genevaTelemetryClient ?? throw new ArgumentException("genevaTelemetryClient cannot be null");
+            _genevaOtelLogger = genevaOpenTelemetryLogger;
```
dgkanatsios, line 26: what are the cases that this is null?

emmayspark: it will be null in prod until we enable Otel in production.

PR #171, opened 2023-10-03
Add SecretDetails to SessionHostStartInfo

jotavalenciano:
To support the deployment of Secrets in the VMs we need this new class in the SessionHostStartInfo

### DIFF
--- a/AgentInterfaces/SecretDetail.cs
+++ b/AgentInterfaces/SecretDetail.cs
@@ -8,7 +8,7 @@ namespace Microsoft.Azure.Gaming.AgentInterfaces
     public class SecretDetail
     {
         /// <summary>
-        /// Identifier secret
+        /// Secret identifier
         /// </summary>
         public string Name { get; set; }
 
@@ -26,4 +26,5 @@ namespace Microsoft.Azure.Gaming.AgentInterfaces
         /// Expiration date of this secret
         /// </summary>
         public string ExpirationDate { get; set; }
+    }
 }
--- a/AgentInterfaces/SecretDetail.cs
+++ b/AgentInterfaces/SecretDetail.cs
@@ -26,10 +26,4 @@ namespace Microsoft.Azure.Gaming.AgentInterfaces
         /// Expiration date of this secret
         /// </summary>
         public string ExpirationDate { get; set; }
-
-        /// <summary>
-        /// Expiration date of the secret.
-        /// </summary>
-        public DateTime ExpirationTimeUtc { get; set; }
-    }
 }
--- a/AgentInterfaces/SecretDetail.cs
+++ b/AgentInterfaces/SecretDetail.cs
@@ -1,0 +1,35 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Azure.Gaming.AgentInterfaces
+{
+    using System;
+
+    public class SecretDetail
+    {
+        /// <summary>
+        /// Identifier secret
+        /// </summary>
+        public string Name { get; set; }
+
+        /// <summary>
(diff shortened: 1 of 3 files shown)


AgentInterfaces/SecretDetail.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        /// <summary>
+        /// The version of this secret in the internal Key Vault
+        /// </summary>
+        public string Version { get; set; }
+
+        /// <summary>
+        /// Expiration date of this secret
+        /// </summary>
+        public string ExpirationDate { get; set; }
```
emmayspark, line 28: why we have 2 different data type for Expiration date?

AgentInterfaces/SessionHostsStartInfo.cs
```diff
@@ -86,6 +86,11 @@ public class SessionHostsStartInfo
         /// </summary>
         public CertificateDetail[] GameCertificates { get; set; }
 
+        /// <summary>
+        /// List of secrets we want to deploy on each container
```
emmayspark, line 90: why on each container? are we planning to support only on container?

AgentInterfaces/SecretDetail.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        /// <summary>
+        /// Expiration date of the secret.
```
emmayspark, line 31: maybe we should add this is for UTC and why UTC type is needed?

AgentInterfaces/SecretDetail.cs
jotavalenciano: I copied the same from CertificateDetail class. I want to make it consistent

AgentInterfaces/SessionHostsStartInfo.cs
jotavalenciano: I copied the same from CertificateDetail class. I want to make it consistent. But your are right. Maybe container is not the right word

AgentInterfaces/SecretDetail.cs
jotavalenciano: I copied the same from CertificateDetail, but you are right. Having both of them is not great

AgentInterfaces/SecretDetail.cs
```diff
@@ -0,0 +1,29 @@
...
+// Licensed under the MIT License.
+
+namespace Microsoft.Azure.Gaming.AgentInterfaces
+{
+    using System;
+
+    public class SecretDetail
+    {
+        /// <summary>
+        /// Identifier secret
```
dgkanatsios, line 11: nit: Secret identifier

AgentInterfaces/SecretDetail.cs
```diff
@@ -0,0 +1,29 @@
...
+        /// <summary>
+        /// The version of this secret in the internal Key Vault
+        /// </summary>
+        public string Version { get; set; }
+
+        /// <summary>
+        /// Expiration date of this secret
+        /// </summary>
+        public string ExpirationDate { get; set; }
+}
```
dgkanatsios, line 29: there's a } missing, as depicted here [https://github.com/PlayFab/MpsAgent/actions/runs/6399628016/job/17371970994?pr=171](https://github.com/PlayFab/MpsAgent/actions/runs/6399628016/job/17371970994?pr=171)

jotavalenciano: OMG yes. Don't know why I missed that

AgentInterfaces/SecretDetail.cs
jotavalenciano: Removing the second parameter

dgkanatsios (approved):
:shipit:

PR #179, opened 2024-03-20
VmAgent endpoint to report GSDK version telemetry

tculotta:
This PR creates an extra endpoint on VmAgent HTTP server to accept environment/version info from GSDK. Implementation is pretty simple since VmAgent will just log the version.

### DIFF
--- a/LocalMultiplayerAgent/Controllers/SessionHostController.cs
+++ b/LocalMultiplayerAgent/Controllers/SessionHostController.cs
@@ -11,7 +11,7 @@ namespace LocalMultiplayerAgent.Controllers
     using Microsoft.Azure.Gaming.LocalMultiplayerAgent;
     using Microsoft.Azure.Gaming.VmAgent.Model;
     using Microsoft.Extensions.Hosting;
-    using Newtonsoft.Json;
+    using Instrumentation;
 
     public class SessionHostController : Controller
     {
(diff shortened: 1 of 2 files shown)


LocalMultiplayerAgent/Controllers/SessionHostController.cs
```diff
@@ -96,6 +96,29 @@ public SessionHostController(IHostApplicationLifetime appLifetime)
...
+                return BadRequest($"{nameof(GsdkVersionInfo.Version)} should not be an empty string");
+            }
+            if (string.IsNullOrEmpty(vi.Flavor))
+            {
+                return BadRequest($"{nameof(GsdkVersionInfo.Flavor)} should not be an empty string");
+            }
+
+            if (!_wasGsdkVersionLogged)
+            {
+                Globals.MultiLogger.LogInformation($"GSDK flavor/version: {vi.Flavor}/{vi.Version} from session host {sessionHostId}");
```
dgkanatsios, line 115: I think we just do Console.WriteLine on LMA, does this work?

tculotta: It does. The multilogger is setup with a console sink.

dgkanatsios (approved):

PR #180, opened 2024-03-27
add configurable maintenance event to local agent

tculotta:
Aligns maintenance schedule to final schema and adds event simulation with json config values (disabled by default).

### DIFF
--- a/AgentInterfaces/MaintenanceSchedule.cs
+++ b/AgentInterfaces/MaintenanceSchedule.cs
@@ -27,15 +27,6 @@ namespace Microsoft.Azure.Gaming.AgentInterfaces
             DocumentIncarnation = other.DocumentIncarnation;
             MaintenanceEvents = other.MaintenanceEvents.Select((e) => new MaintenanceEvent(e)).ToList();
         }
-
-        public MaintenanceSchedule(string eventType, string eventStatus, string eventSource)
-        {
-            DocumentIncarnation = "1";
-            MaintenanceEvents = new List<MaintenanceEvent>()
-            {
-                new MaintenanceEvent(eventType, eventStatus, eventSource)
-            };
-        }
     }
 
     // https://docs.microsoft.com/en-us/azure/virtual-machines/windows/scheduled-events#query-for-events
@@ -48,7 +39,8 @@ namespace Microsoft.Azure.Gaming.AgentInterfaces
 
         public string ResourceType { get; set; }
 
-        public IList<string> Resources { get; set; }
+        [JsonProperty("Resources")]
+        public IList<string> AffectedResources { get; set; }
 
         public string EventStatus { get; set; }
 
@@ -70,24 +62,11 @@ namespace Microsoft.Azure.Gaming.AgentInterfaces
             EventId = other.EventId;
             EventType = other.EventType;
             ResourceType = other.ResourceType;
-            Resources = other.Resources.ToList();
+            AffectedResources = other.AffectedResources.ToList();
             EventStatus = other.EventStatus;
             NotBefore = other.NotBefore;
             EventSource = other.EventSource;
             DurationInSeconds = other.DurationInSeconds;
-        }
-
-        public MaintenanceEvent(string eventType, string eventStatus, string eventSource)
-        {
-            EventId = Guid.NewGuid().ToString();
-            EventType = eventType;
-            ResourceType = "VirtualMachine";
-            Resources = new List<string>() { "vmId" };
-            EventStatus = eventStatus;
-            NotBefore = DateTime.UtcNow.AddMinutes(5);
-            Description = $"Scheduled {eventType} event for VM";
-            EventSource = eventSource;
-            DurationInSeconds = 300;
         }
     }
 }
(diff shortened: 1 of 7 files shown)


AgentInterfaces/MaintenanceSchedule.cs
```diff
@@ -43,13 +48,14 @@ public class MaintenanceEvent
 
         public string ResourceType { get; set; }
 
-        [JsonProperty("Resources")]
-        public IList<string> AffectedResources { get; set; }
```
dgkanatsios, line 47: we should not change property names, these classes can (and are) used in a number of places.

AgentInterfaces/MaintenanceSchedule.cs
```diff
@@ -64,11 +70,24 @@ public MaintenanceEvent(MaintenanceEvent other)
...
             ResourceType = other.ResourceType;
-            AffectedResources = other.AffectedResources.ToList();
+            Resources = other.Resources.ToList();
             EventStatus = other.EventStatus;
             NotBefore = other.NotBefore;
             EventSource = other.EventSource;
             DurationInSeconds = other.DurationInSeconds;
         }
+
+        public MaintenanceEvent(string eventType, string eventStatus, string eventSource)
```
dgkanatsios, line 80: is this constructor used in testing? If so, it should not be in this class but as a factory method in the test project.

AgentInterfaces/MaintenanceSchedule.cs
```diff
@@ -29,7 +26,15 @@ public MaintenanceSchedule(MaintenanceSchedule other)
         {
             DocumentIncarnation = other.DocumentIncarnation;
             MaintenanceEvents = other.MaintenanceEvents.Select((e) => new MaintenanceEvent(e)).ToList();
-            ReportingVmId = other.ReportingVmId;
+        }
+
+        public MaintenanceSchedule(string eventType, string eventStatus, string eventSource)
```
dgkanatsios, line 31: similar as the other comment, if this constructor is used in testing, code should reside in the test project.

LocalMultiplayerAgent/Controllers/SessionHostController.cs
```diff
@@ -78,22 +79,32 @@ public SessionHostController(IHostApplicationLifetime appLifetime)
                     config = Globals.SessionConfig;
                 }
 
+                sendMaintenanceEvent = (numHeartBeats == Globals.Settings.NumHeartBeatsForMaintenanceEventResponse);
+
                 HeartBeatsCount[sessionHostId]++;
             }
             else
             {
+                sendMaintenanceEvent = (numHeartBeats == 1);
```
dgkanatsios, line 88: why are we doing this here?

LocalMultiplayerAgent/MultiplayerSettings.json
```diff
@@ -2,8 +2,13 @@
     "RunContainer": false,
     "OutputFolder": "",
     "NumHeartBeatsForActivateResponse": 10,
+    "NumHeartBeatsForMaintenanceEventResponse": 0, // a value < 1 will disable the maintenance event
```
dgkanatsios, line 5: I'd say we start with a negative number, to disable this by default, since that's the way LMA works now (not sending maintenance events)

dgkanatsios (changes requested):
Comments submitted, also tests are failing since projects don't build

AgentInterfaces/MaintenanceSchedule.cs
tculotta: Didn't realize this project was pulled into the production agent. I'll revert this and move other changes out of here

LocalMultiplayerAgent/MultiplayerSettings.json
tculotta: 0 is disabled already. Heartbeat count starts at 1

LocalMultiplayerAgent/Controllers/SessionHostController.cs
tculotta: Changing this. Should've been checking NumHeartBeatsForMaintenanceEventResponse

AgentInterfaces/MaintenanceSchedule.cs
```diff
@@ -15,11 +15,9 @@ public class MaintenanceSchedule
     {
         public string DocumentIncarnation { get; set; }
 
-        [JsonProperty("Events")]
+        [JsonProperty("events")]
         public IList<MaintenanceEvent> MaintenanceEvents { get; set; }
 
-        public string ReportingVmId { get; set; }
```
dgkanatsios, line 21: why are we removing this?

LocalMultiplayerAgent.UnitTest/ControllerTests.cs
```diff
@@ -30,6 +30,8 @@ public async Task ValidateStandingByToInitializingThrows()
 
             string sessionHostId = "testSessionHostId";
 
+            Globals.Settings = new();
```
dgkanatsios, line 33: can we add a comment on why we're doing this?

AgentInterfaces/MaintenanceSchedule.cs
tculotta: That's not part of the schema

dgkanatsios (approved):

PR #181, opened 2024-04-01
Update Windows os image

valexao:
updated image tag for windows os image


dgkanatsios (approved):

dgkanatsios (changes requested):
Do we need this extra file?

valexao:
not sure what this file is but looks like it will not be checked in? seems like some github beta view thing 
image

dgkanatsios:
Looks like the PR has two files, we should remove the two in the "LocalMultiplayerAgentPublishFolder"

dgkanatsios (approved):

PR #182, opened 2024-04-10
Add MPS Secrets as Env Variables in Game Servers

jotavalenciano:
Adding the MPS Secrets as Environment Variables in Game Servers. They will also be accessed by the VMStartupScript once we update the nuget package in VMAgent


dgkanatsios (approved):
:shipit:

PR #191, opened 2025-04-09
adding a ToRedactedString method

brianwp3000:
Trying to remove a bunch of text from VmPersistedState to make it easier to fully redact later. I just went through a bunch of our model classes and add a ToRedacted method, which clones the object and nulls out text we want to remove.
I tried being fancy with Newtonsoft, but i couldn't figure out a way to do it without messing with places where we don't want to redact this text. I think i could have done it with system.text but we're not using that everywhere

### DIFF
--- a/AgentInterfaces/AssetDetail.cs
+++ b/AgentInterfaces/AssetDetail.cs
@@ -22,5 +22,12 @@ namespace Microsoft.Azure.Gaming.AgentInterfaces
         /// For developer testing, specify a file path accessible locally.
         /// </summary>
         public string LocalFilePath { get; set; }
+
+        public AssetDetail ToRedacted()
+        {
+            AssetDetail clone = (AssetDetail)this.MemberwiseClone();
+            clone.SasTokens = null;
+            return clone;
+        }
     }
 }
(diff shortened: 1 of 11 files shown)


AgentInterfaces/AssetDetail.cs
```diff
@@ -22,5 +22,12 @@ public class AssetDetail
         /// For developer testing, specify a file path accessible locally.
         /// </summary>
         public string LocalFilePath { get; set; }
+
+        public AssetDetail ToRedacted()
+        {
+            AssetDetail clone = (AssetDetail)this.MemberwiseClone();
```
dgkanatsios, line 28: how can we verify that we don't forget to redact a property we add later? any chance we can have a test to verify that?

brianwp3000: I don't think adding a test for this would be possible--there's no way to automate knowing what fields need to be redacted. Not even an automated credential scanning library could do that in a unit test, because the library wouldn't flag the new property it unless we filled it with an actual credential

dgkanatsios (approved):
:shipit:

PR #194, opened 2025-04-18
Removing Serializable attribute

dgkanatsios:
This PR removes the legacy serializable attribute which we don't need any more

### DIFF
--- a/VmAgent.Core/VmDirectories.cs
+++ b/VmAgent.Core/VmDirectories.cs
@@ -41,9 +41,9 @@ namespace Microsoft.Azure.Gaming.VmAgent.Core
 
             AgentStateFile = Path.Combine(PlayFabFolderOnPrimaryDrive, "PlayFabVmAgentState");
             AgentStateTempFile = Path.Combine(PlayFabFolderOnPrimaryDrive, "PlayFabVmAgentState.tmp");
-            MonitoringStateFile = Path.Combine(TempStorageRootVm, "PlayFabMonitoringState");
-            MonitoringStateTempFile = Path.Combine(TempStorageRootVm, "PlayFabMonitoringState.tmp");
-            HostConfigOverrideFile = Path.Combine(TempStorageRootVm, "HostConfigOverride.json");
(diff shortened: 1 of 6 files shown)


VmAgent.Core/VmDirectories.cs
```diff
@@ -41,9 +41,9 @@ public VmDirectories(string rootPath)
 
             AgentStateFile = Path.Combine(PlayFabFolderOnPrimaryDrive, "PlayFabVmAgentState");
             AgentStateTempFile = Path.Combine(PlayFabFolderOnPrimaryDrive, "PlayFabVmAgentState.tmp");
-            MonitoringStateFile = Path.Combine(TempStorageRootVm, "PlayFabMonitoringState");
-            MonitoringStateTempFile = Path.Combine(TempStorageRootVm, "PlayFabMonitoringState.tmp");
-            HostConfigOverrideFile = Path.Combine(TempStorageRootVm, "HostConfigOverride.json");
+            MonitoringStateFile = Path.Combine(PlayFabFolderOnPrimaryDrive, "PlayFabMonitoringState");
+            MonitoringStateTempFile = Path.Combine(PlayFabFolderOnPrimaryDrive, "PlayFabMonitoringState.tmp");
```
brianwp3000, line 45: Why are these files being moved out of the temporary folder? Especially this .tmp file--i would expect that to be in a temporary folder

brianwp3000 (changes requested):

VmAgent.Core/VmDirectories.cs
dgkanatsios: wrong commit included by mistake, removed, thanks!

brianwp3000 (approved):

brianwp3000:
Also, last time i made a change to AgentInterfaces i ended up going back and bumping the VersionNumber in the csproj because our docs say that's necessary to publish the nuget package--i actually don't think that's necessary anymore. The new nuget publish pipeline uses the current date as the nuget package version

